PR #1, opened 2021-09-28
feat: Add an RST-style docstring parser

pawamoy:
@plannigan, since you originally added the parser in pytkdocs, maybe you'd like to review this refactored version. Also let me know what you think is the most appropriate way to give you credits :slightly_smiling_face:  It can be as a co-author of this PR / these commits, as a comment in the module, in the module docstring, or in the repository credits, or all of these! I'll anyway add every contributor of pytkdocs to the repository credits.

A quick explanation of the changes: there's no base parser anymore, so I exploded the class into several functions. They pass the `parsed_values` around to modify it by side-effect (not sure this is the right approach now that we're not using a class). The lines can be obtained from the docstring object, which has access to the whole object tree through its parent attribute. This removes the need for the obscure "context" dictionary we had in pytkdocs.

As for what is this project: it's a revisited pytkdocs, which parses the code instead of executing it (loading it in memory to introspect objects). It's far more robust and performant, and will allow extensions to customize the behavior.

### DIFF
--- a/config/flake8.ini
+++ b/config/flake8.ini
@@ -119,4 +119,4 @@ per-file-ignores =
     src/griffe/dataclasses.py:WPS115
     src/griffe/visitor.py:N802,D102
     src/griffe/encoders.py:WPS232
-    tests/*:WPS118,WPS218
+    tests/*:WPS116,WPS118,WPS218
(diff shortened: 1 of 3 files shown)


pawamoy:
I went with credits in the module docstring and co-authoring!

PR #56, opened 2022-03-05
feat: Ignore `__doc__` from parent classes

### LINKED ISSUE #55: Ignore `__doc__` from parent classes
Currently Griffe will fall-back to extracting docstrings from the `__doc__` attribute if it cannot obtain them from the source file. When it does so, it will make use of the `__doc__` attribute of parent classes if the derived class does not have a docstring. This frequently leads to misleading documentation.

For example, if you create a class that inherits from `numbers.Real`, your class will have the following text picked up by Griffe unless you define a docstring:

```
>>> print(numbers.Real.__doc__)
To Complex, Real adds the operations that work on real numbers.

    In short, those are: a conversion to float, trunc(), divmod,
    %, <, <=, >, and >=.

    Real also provides defaults for the derived operations.
```

Instead of using `getattr(x, '__doc__')`, Griffe can directly access the members of the class it is gathering docstrings from with `x.__dict__['__doc__']`.

This would make the behaviour more consistent with how Griffe extracts docstrings from regular Python modules, as for those it does not use the docstring from parent classes at all.

WillDaSilva:
Closes  #55.

### DIFF
--- a/src/griffe/agents/inspector.py
+++ b/src/griffe/agents/inspector.py
@@ -184,12 +184,10 @@ class Inspector(BaseInspector):  # noqa: WPS338
         value = getattr(node.obj, "__doc__", None)
         if value is None:
             return None
-        else:
+        return Docstring(
             # `inspect.getdoc` calls `inspect.cleandoc`. We avoid `inspect.getdoc` to avoid taking
             # the `__doc__` attribute from a parent class, but we still want to clean the doc.
-            value = cleandoc(value)
-        return Docstring(
-            value,
+            cleandoc(value),
             parser=self.docstring_parser,
             parser_options=self.docstring_options,
         )
--- a/src/griffe/agents/inspector.py
+++ b/src/griffe/agents/inspector.py
@@ -25,7 +25,7 @@ from __future__ import annotations
 import ast
 import sys
 from inspect import Parameter as SignatureParameter
-from inspect import Signature, getdoc, getmodule, ismodule
+from inspect import Signature, cleandoc, getmodule, ismodule
 from inspect import signature as getsignature
 from pathlib import Path
 from tokenize import TokenError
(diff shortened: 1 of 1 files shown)


src/griffe/agents/inspector.py
```diff
@@ -180,9 +180,14 @@ def __init__(
         self.lines_collection: LinesCollection = lines_collection or LinesCollection()
 
     def _get_docstring(self, node: ObjectNode) -> Docstring | None:
-        value = getdoc(node.obj)
+        # Access `__doc__` directly to avoid taking the `__doc__` attribute from a parent class.
+        value = getattr(node.obj, "__doc__", None)
         if value is None:
             return None
+        else:
```
pawamoy, line 187: Useless `else`!

WillDaSilva: Good catch.

WillDaSilva:
Seems like some or all of the CI failures are occurring on `master` too, and so are unrelated to this change.

1. `make format` works.
2. `make check` fails on `master` and on these changes with the same error:
    ```
    TypeError: __init__() missing 1 required positional argument: 'metrics'
    ```
4. `make test` runs all of the tests without failure:
    ```
    will ~/workshop/griffe (master) $ make test
    > Currently selected Python version: python3
    > pdm run duty test (python3.7)
    ✓ Running tests
    > pdm run duty test (python3.8)
    ✓ Running tests
    > pdm run duty test (python3.9)
    ✓ Running tests
    > pdm run duty test (python3.10)
    ✓ Running tests
    > pdm run duty test (python3.11)
    ✓ Running tests
    > Restored previous Python version: python3
    ```

pawamoy (approved):
LGTM, thanks a lot!

pawamoy:
Yeah don't worry about CI :slightly_smiling_face:

Issue #59, opened 2022-03-21
Default value not extracted when coming from a dict


Thomzoy:
**Describe the bug**
When defining  default values for parameters of a function from a dictionary, *Griffe* doesn't extract this default value:

**To Reproduce**
Project structure:
```
test_griffe
│   __init__.py
│   my_mod.py  
```

In `my_mod.py`:
```python
DEFAULT_DICT = dict(
    key1="value1",
)

DEFAULT_VALUE = 3


def f(
    param1: str = DEFAULT_DICT["key1"],
    param2: int = DEFAULT_VALUE,
):
    """
    A griffe test function

    Parameters
    ----------
    param1 : str, optional
        _description_, by default DEFAULT_DICT["key1"]
    param2 : int, optional
        _description_, by default DEFAULT_VALUE
    """
    return
```

Truncated output of `griffe test_griffe`
```json
          {
            "kind": "function",
            "name": "f",
            "lineno": 9,
            "docstring": {
              "value": "[THE DOCSTRING]",
              "lineno": 25,
              "endlineno": null
            },
            "labels": [],
            "members": [],
            "decorators": [],
            "parameters": [
              {
                "name": "param1",
                "annotation": {
                  "source": "str",
                  "full": "str"
                },
                "kind": "positional or keyword",
                "default": null
              },
              {
                "name": "param2",
                "annotation": {
                  "source": "int",
                  "full": "int"
                },
                "kind": "positional or keyword",
                "default": "DEFAULT_VALUE"
              }
            ],
            "returns": null
          }
```
**Expected behavior**
- The `"default"` for param2 is correct (although a side question might be, could it display `3`, the actual value stored in `DEFAULT_VALUE` ?)
- The default for `param1` is `null`, while I was expecting one of the following:
   - "DEFAULT_DICT['key1']
   - "value1"

Maybe I'm just not fiddling with the options enough ?

Thnaks for your help !

**System (please complete the following information):**
- `griffe` version: [0.14.1]
- Python version: [3.7]
- OS: [Linux]

pawamoy:
Hey again @Thomzoy, thanks for the report!

> 
    The "default" for param2 is correct (although a side question might be, could it display 3, the actual value stored in DEFAULT_VALUE ?)

[...]

PR #60, opened 2022-03-21
fix: retrieve default value for non-string parameters

Thomzoy:
Should solve #59 and [this issue](https://github.com/mkdocstrings/python/issues/8) from `mkdocstrings-python`

### DIFF
--- a/src/griffe/agents/nodes.py
+++ b/src/griffe/agents/nodes.py
@@ -535,6 +535,7 @@ if sys.version_info < (3, 8):
         """
         return {elt.s for elt in node.value.elts}  # type: ignore[attr-defined]
 
+
 else:
 
     def parse__all__(node: NodeAssign) -> set[str]:  # noqa: WPS116,WPS120,WPS440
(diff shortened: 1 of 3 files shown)


src/griffe/agents/nodes.py
```diff
@@ -535,6 +535,7 @@ def parse__all__(node: NodeAssign) -> set[str]:  # noqa: WPS116,WPS120
         """
         return {elt.s for elt in node.value.elts}  # type: ignore[attr-defined]
 
+
```
pawamoy, line 538: Useless change (and Black would not be happy), please remove :smile: 

```suggestion
```

src/griffe/agents/nodes.py
```diff
@@ -1186,12 +1187,9 @@ def get_parameter_default(node: AST, filepath: Path, lines_collection: LinesColl
...
         return None
-    if isinstance(node, NodeConstant):
-        return repr(node.value)
-    if isinstance(node, NodeStr):
-        return repr(node.s)
-    if isinstance(node, NodeName):
-        return node.id
+    value_to_str = _node_value_map.get(type(node), None)
+    if value_to_str is not None:
+        return value_to_str(node)
```
pawamoy, line 1192: Something simpler:

```suggestion
    with suppress(KeyError):
        return get_value(node)
```

And `from contextlib import suppress` must be added to the module imports :slightly_smiling_face:

pawamoy (changes requested):
Thanks a lot!

Could you also add tests for this? You can get inspiration from `test_building_annotations_from_nodes` in `test_nodes.py`. Let me know if you need more hints.

pawamoy:
We should test simple defaults like strings, integers, floats and booleans, but also more complex expressions like `DEFAULT_DICT["key1"]`, as you showed in #59 :slightly_smiling_face:

src/griffe/agents/nodes.py
Thomzoy: Actually black added this line !
I have version 21.9b

src/griffe/agents/nodes.py
Thomzoy: Good one !

src/griffe/agents/nodes.py
pawamoy: I see, my bad!

pawamoy (approved):
LGTM (CI good enough, will fix in other commits)!

Thanks again!

PR #61, opened 2022-04-03
tests: Fix CLI Test on Windows

bandersen23:
Changed the output path of CLI test on Windows to a similar path. 

This should fix the Windows build error as I patched it similarly here on the `conda-forge` Windows test 👍 : https://github.com/conda-forge/griffe-feedstock/blob/befe014bfac10bb45a46ad7e19d598c34e53db99/recipe/patches/0001-windows-cli-test-patch.patch

> Reference: 
> https://stackoverflow.com/questions/313111/is-there-a-dev-null-on-windows
> https://stackoverflow.com/questions/446209/possible-values-from-sys-platform

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -9,7 +9,7 @@ from griffe import cli
 
 def test_main():
     """Basic CLI test."""
-    if sys.platform == 'windows':
+    if sys.platform == 'win32':
         assert cli.main(["griffe", "-s", "src", "-oNUL"]) == 0
     else:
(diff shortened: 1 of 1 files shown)


pawamoy:
I ran `make format`, Black was complaining :slightly_smiling_face:

bandersen23:
Awesome! I did this on the GitHub website because I didn't want to forget about it, so I couldn't do the `make format` command 😄

pawamoy:
Thanks a  lot ^^ The last error is related to mypy, not sure why it fails in CI while working locally :/ Unrelated to this PR anyway, so I'll merge :)

PR #63, opened 2022-04-08
feat: warn if docstring has extra parameters

KevinMusgrave:
Reference: https://github.com/mkdocstrings/mkdocstrings/issues/408

### DIFF
--- a/src/griffe/dataclasses.py
+++ b/src/griffe/dataclasses.py
@@ -247,6 +247,9 @@ class Parameters:
     def __iter__(self):
         return iter(self._parameters_list)
 
+    def __contains__(self, item):
+        return item in self._parameters_dict
+
     def add(self, parameter: Parameter) -> None:
         """Add a parameter to the container.
 
--- a/src/griffe/dataclasses.py
+++ b/src/griffe/dataclasses.py
@@ -247,8 +247,8 @@ class Parameters:
     def __iter__(self):
         return iter(self._parameters_list)
 
-    def __contains__(self, item):
-        return item in self._parameters_dict
+    def __contains__(self, param_name: str):
+        return param_name in self._parameters_dict
 
     def add(self, parameter: Parameter) -> None:
         """Add a parameter to the container.
--- a/src/griffe/docstrings/google.py
+++ b/src/griffe/docstrings/google.py
@@ -209,7 +209,7 @@ def _read_parameters(
                 _warn(
                     docstring,
                     line_number,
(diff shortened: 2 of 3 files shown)


src/griffe/dataclasses.py
```diff
@@ -247,6 +247,9 @@ def __len__(self):
     def __iter__(self):
         return iter(self._parameters_list)
 
+    def __contains__(self, item):
+        return item in self._parameters_dict
```
pawamoy, line 251:
```suggestion
    def __contains__(self, param_name: str):
        return param_name in self._parameters_dict
```

src/griffe/docstrings/google.py
```diff
@@ -202,6 +204,14 @@ def _read_parameters(docstring: Docstring, offset: int) -> tuple[list[DocstringP
         if annotation is None:
             _warn(docstring, line_number, f"No type or annotation for parameter '{name}'")
 
+        if warn_unknown_params and docstring.parent is not None:
+            if name not in docstring.parent.parameters:
+                _warn(
+                    docstring,
+                    line_number,
+                    f'The docstring for {docstring.parent} contains "{name}" as a parameter, but "{name}" is not an argument.',
```
pawamoy, line 212:
```suggestion
                    f"Parameter '{name}' does not appear in the parent signature",
```

tests/test_docstrings/test_google.py
```diff
@@ -702,22 +702,34 @@ def test_parameter_line_without_colon_keyword_only(parse_google):
...
+                Parameter("y"),
+            ),
+        ),
+    )
+    assert len(sections) == 1
+    assert len(warnings) == 3
+    assert (
+        'The docstring for <Function(\'func\', None, None)> contains "x" as a parameter, but "x" is not an argument.'
+        in warnings
+    )
```
pawamoy, line 732:
```suggestion
    assert "'x' does not appear in parent signature" in warnings[1]
```

pawamoy (changes requested):
Hey, thanks a lot!

A few suggestions. We also must pass `warn_unknown_params=False` from `_read_other_parameters`, and add a test to make sure "Other Parameters" sections will not emit warnings.

KevinMusgrave:
I added ```warn_unknown_params=False``` and tested for it in ```test_extra_parameter```

pawamoy:
I've split the test in two, to make it more resilient to future possible changes in the warning string: we just check no warnings were emitted.

pawamoy (approved):
LGTM

pawamoy:
Thanks again!

PR #65, opened 2022-04-18
feat: Inherit class parameters from __init__

francois-rozet:
See https://github.com/mkdocstrings/python/issues/19.

### DIFF
--- a/src/griffe/dataclasses.py
+++ b/src/griffe/dataclasses.py
@@ -1131,9 +1131,15 @@ class Class(Object):
 
     @property
     def parameters(self) -> Parameters:
+        """"Return the parameters of this class' __init__ method, if any.
+
+        Returns:
+            The parameters containter.
+        """
+
         try:
             return self.members["__init__"].parameters
-        except:
+        except KeyError:
             return Parameters()
 
     def as_dict(self, **kwargs: Any) -> dict[str, Any]:  # type: ignore[override]
--- a/src/griffe/dataclasses.py
+++ b/src/griffe/dataclasses.py
@@ -1131,14 +1131,13 @@ class Class(Object):
 
     @property
     def parameters(self) -> Parameters:
-        """"Return the parameters of this class' __init__ method, if any.
+        """Return the parameters of this class' `__init__` method, if any.
 
         Returns:
             The parameters containter.
         """
-
         try:
-            return self.members["__init__"].parameters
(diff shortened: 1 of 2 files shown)


src/griffe/dataclasses.py
```diff
@@ -1129,6 +1129,13 @@ def __init__(
         self.bases: list[Name | Expression | str] = bases or []
         self.decorators: list[Decorator] = decorators or []
 
+    @property
+    def parameters(self) -> Parameters:
+        try:
+            return self.members["__init__"].parameters
+        except:
```
pawamoy, line 1136:
```suggestion
        except KeyError:
```

pawamoy (changes requested):
Thanks! Could you also add a docstring to the property, and a test that makes sure parameters section in docstrings of classes with init methods can be parsed? You can add it right after `test_unknown_params_scan_doesnt_crash_on_non_function_docstrings` in `test_google.py`.
I can give more details if needed.

francois-rozet:
Yep I'll do this shortly. Should the test be only for the Google style?

Also, you might actually want to revert the changes from https://github.com/mkdocstrings/griffe/commit/4a417bc6c0e83b42fe1a74a4a8b0881d3955075f, as they would not be necessary anymore.

pawamoy:
I'll not revert them but rephrase the comment and change the test indeed, as we still don't want to crash if a module or an attribute has a parameters section (who knows! could happen).

francois-rozet:
I did the modifications and added the new test. However, I was not able to run the tests unfortunately as `make setup` doesn't work on my machine.

pawamoy:
Any error message so I can improve the setup script? Did you try to install PDM and dependencies as described in [Contributing](https://mkdocstrings.github.io/griffe/contributing/#environment-setup)?

pawamoy (approved):
OK, looking good! Thanks!

francois-rozet:
> Any error message so I can improve the setup script?

```
$ make setup
  installed package pdm 1.14.0, installed using Python 3.8.10
  These apps are now globally available
    - pdm
done! ✨ 🌟 ✨
[KeyError]: 'python.path'
Add '-v' to see the detailed traceback
make: *** [Makefile:42: setup] Error 1
```

pawamoy:
Ah, I see, thanks, I'll be able to fix it :slightly_smiling_face:

PR #74, opened 2022-05-29
feat: add to/from_json convenience methods to Object

tlambert03:
👋

I wonder if you'd consider this PR that adds some serialization/deserialization convenience methods to `dataclasses.Object`.  When using programmatically, it makes it easier to construct/deconstruct from serialized.

I made it a mixin that could be added to other objects (since _many_ of the objects here could theoretically be supported), but I stopped short of adding it to anything besides Object.  (but `ModulesCollection`, for example, would be another prime candidate).

One of the nice things this does is support type hinting:
```python
from griffe.dataclasses import Module

mod = Module.from_json('{...}')
reveal_type(mod)  # type of "mod" is "Module"
```

### DIFF
--- a/src/griffe/mixins.py
+++ b/src/griffe/mixins.py
@@ -138,12 +138,11 @@ class ObjectAliasMixin:
 class SerializationMixin:
     """A mixin that adds de/serialization conveniences."""
 
-    def as_json(self, full: bool = False, indent: int = 2, **kwargs: Any) -> str:
+    def as_json(self, full: bool = False, **kwargs: Any) -> str:
         """Return this object's data as a JSON string.
 
         Parameters:
             full: Whether to return full info, or just base info.
-            indent: Indentation level for JSON output.
             **kwargs: Additional serialization options passed to encoder.
 
         Returns:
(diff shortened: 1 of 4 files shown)


pawamoy:
Thanks! That looks very convenient indeed. I just feel like the test you added is very redundant. Maybe you could simply rewrite the end-to-end test using the new `as_json` and `from_json` methods, so we get coverage for everything, and keep the part that checks we can't load a module as a function.

src/griffe/mixins.py
```diff
@@ -131,3 +133,45 @@ def is_implicitely_exported(self) -> bool:
...
+        Parameters:
+            json_string: JSON to decode into Object.
+            **kwargs: Additional options passed to decoder.
+
+        Returns:
+            An Object instance.
+
+        Raises:
+            TypeError: When the json_string does not represent and object
+                       of the class from which this classmethod has been called.
```
pawamoy, line 169: Doesn't the Google parser complain about the indentation here :sweat_smile:? It should be indented once more, not aligned with the beginning of the sentence.

tlambert03: lol.  yeah, I did that trying to satisfy the parser :joy: ... will tone it back a bit :)

tlambert03:
all set 👍

tlambert03:
i think the remaining ci / quality fails are the same as in master: https://github.com/mkdocstrings/griffe/runs/6644495952?check_suite_focus=true#step:10:16

pawamoy:
If it's coming from master then lets leave it like that. I'll fix it later. LGTM!

pawamoy:
Thanks again!!

PR #76, opened 2022-05-29
feat: add load_git, to load Module from a git committish

tlambert03:
Since you mentioned liking the module-from-git approach in griffediff.  Here's a PR for your consideration that adds it here.  This will create a temporary [git worktree](https://git-scm.com/docs/git-worktree) with a specific commit or tag checked out, then add it to the search_paths passed to the loader

```py
from griffe.git import load_git

module = load_git('my_module', commit='v0.1.0', repo='path/to/repository')
```

~leaving in draft for now, as i'm sure there's some stuff to discuss (and I haven't run all the checks yet).~


tlambert03:
Fixed up docstrings and checks... and created a better test fixture strategy.

Note, I had to `noqa` a bunch of `subprocess` warnings (S404,S603,S607) if that concerns you

pawamoy:
Thanks, I didn't get the time to review yet, but soon :slightly_smiling_face:

config/mypy.ini
```diff
@@ -1,6 +1,6 @@
 [mypy]
 ignore_missing_imports = true
-exclude = tests/fixtures/
+exclude = tests/fixtures/|tests/test_git/_repo
```
pawamoy, line 3: The `fixtures` directory is exactly made for this! Can you move `_repo` inside it?
You have some `Path` instances already defined in `tests/__init__.py` that you can import:

```python
from tests import FIXTURES_DIR
```

src/griffe/git.py
```diff
@@ -0,0 +1,120 @@
...
+from griffe.dataclasses import Module
+from griffe.docstrings.parsers import Parser
+
+
+def _assert_git_repo(repo: str) -> None:
+    if not shutil.which("git"):
+        raise RuntimeError("Could not find git executable. Please install git.")
+
+    try:
+        check_output(["git", "-C", repo, "rev-parse", "--is-inside-work-tree"], stderr=DEVNULL)  # noqa: S603,S607
```
pawamoy, line 33: I think you can use `run` with `check=True` instead of `check_output`, that would be one less import. What do you think? Maybe I'm missing something.

src/griffe/git.py
```diff
@@ -0,0 +1,120 @@
...
+        The path to the temporary worktree.
+
+    Raises:
+        OSError: If `repo` is not a valid `.git` repository
+        RuntimeError: If the `git` executable is unavailable, or if it cannot create a worktree
+    """
+    repo = str(repo)
+    _assert_git_repo(repo)
+    with TemporaryDirectory() as td:
+        uid = f"{str(uuid1())[:5]}_{commit}".replace(" ", "_")
```
pawamoy, line 56: Is there a reason why you chose UUID1 instead of UUID4 for example? And why you keep only the first 5 characters? Also, I don't see how spaces could be found in the result :confused: UUID1 and commit sha are only composed of `[0-9a-f]`, no?

src/griffe/git.py
```diff
@@ -0,0 +1,120 @@
...
+        lines_collection: A collection of source code lines.
+        modules_collection: A collection of modules.
+        allow_inspection: Whether to allow inspecting modules when visiting them is not possible.
+
+    Returns:
+        A loaded module.
+    """
+    kwargs = locals().copy()
+    kwargs.pop("commit")
+    kwargs.pop("repo")
```
pawamoy, line 115: I know this is convenient but I'd prefer the verbose, explicit counterpart where you pass each arguments to `loader.load` again :sweat_smile:

tests/test_git/test_from_git.py
```diff
@@ -0,0 +1,93 @@
+"""Tests for creating a griffe Module from specific commits in a git repository."""
```
pawamoy, line 1: Since there's only one test file for now (and if you move the `_repo` to the `fixtures` directory as mentioned above), you can move this up one directory, and rename it `test_git.py` since that's the module name in Griffe sources.

tests/test_git/test_from_git.py
```diff
@@ -0,0 +1,93 @@
...
+from pathlib import Path
+from subprocess import run  # noqa: S404
+
+import pytest
+
+from griffe.dataclasses import Module
+from griffe.git import load_git
+
+REPO_NAME = "my-repo"
+REPO_SOURCE = Path(__file__).parent / "_repo"
```
pawamoy, line 12: This is where you can use `from tests import FIXTURES_DIR` :smile:

pawamoy (changes requested):
Review done! Mostly good, just one or two questions, and some nitpicks about file structure :slightly_smiling_face:

config/mypy.ini
tlambert03: sure. there is no `test/fixtures` folder.  but I can create one

src/griffe/git.py
tlambert03: really, there's not much need for a uuid at all :joy: The commit-ish should be sufficient.  I removed uuid and prefixed instead with `griffe_` ... to indicate who made the worktree (on the very little chance that the cleanup after the `finally` fails

tlambert03:
all set

src/griffe/git.py
pawamoy: Great! That simplifies a lot :slightly_smiling_face: Thanks

pawamoy (approved):
OK! Thanks a lot! This paves the way to great things :sunglasses:

PR #78, opened 2022-05-30
fix: fix stub merging for single-file packages

### LINKED ISSUE #77: Feature: support type stubs?
As mentioned in https://github.com/mkdocstrings/griffe/issues/75#issuecomment-1140480979, one benefit of static API analysis is the ability to pick up stuff like `@typing.overload`-decorated functions, and other stuff that doesn't make it into a runtime representation.

This got me thinking: would it be within scope to support type stub / `.pyi` files?

as an example, consider this terrible, awful, very bad module:
```python
# mymodule.py
globals()['f'] = lambda x: str(x)
```

Yes, this is a silly thing to do, but there are _some_ cases where meta-programming is particularly powerful in python, so it's not out of the realm of "reasonable" possibilities to inject names into a namespace dynamically.  If I'm ever tempted to do something like that, I'd always provide a type stub:

```python
# mymodule.pyi
def f(x: int) -> str: ...
```

currently, griffe would find nothing:
```python
from griffe import load
mod = load('mymodule')
print(mod.members)  # {}
```

... but perhaps it could also look for type stubs to find that info?  Whether a type stub should simply override an actual `py` module, or whether there should be some attempt to merge is another question.  (I'd tend to simply override if present, as type checkers seem to do)

tlambert03:
closes #77 

`pyi` stub files were not being detected and merged in two cases:
1. single-module packages, where a `.pyi` file is also distributed and present in site-packages
2. top module `__init__.pyi`.  This is probably more common than the first case.


src/griffe/loader.py
```diff
@@ -335,6 +337,34 @@ def _store_and_return(self, name: str, module: Module) -> Module:
         self.modules_collection[module.path] = module
         return self.modules_collection[name]  # type: ignore[index]
 
+    def _load_package(self, pkg: Package, submodules: bool = True) -> Module:
+        """Load top module in a package.
```
tlambert03, line 341: not sure if you want this docstring there or not.  Can remove to keep consistent with other privates here?

src/griffe/finder.py
```diff
@@ -137,13 +139,17 @@ def find_module(self, module_name: str) -> Path | list[Path]:  # noqa: WPS231
                     abs_path = path / choice
                     if abs_path in path_contents:
                         if abs_path.suffix:
-                            return abs_path
+                            found_paths.append(abs_path)
```
tlambert03, line 142: 👀 : this of course comes at the loss of early exiting.  So could perhaps slow things down with large search paths.  If you'd prefer that we only look for a pyi file once the original `py` file has been found, let me know

src/griffe/loader.py
pawamoy: Yeah if it's private you can remove the docstring :smile:

pawamoy (changes requested):

.gitignore
```diff
@@ -17,3 +17,4 @@ __pypackages__/
 profile.pstats
 profile.svg
 .hypothesis/
+.vscode/
```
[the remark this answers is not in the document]
tlambert03: Can do!  I saw the .idea folder in there so assumed you were doing the local user-specific thing here :)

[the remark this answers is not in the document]
pawamoy: True haha. You can leave it, it won't hurt :slightly_smiling_face:

src/griffe/finder.py
tlambert03: what about `.pyi` files that are there for non `.py` files, like `.py[cod]`, and `.so` files which we're currently skipping?

tlambert03:
ok, I moved the stub finding logic outside the finder.  It will continue to fail to find stubs for which an actual `.py` file doesn't exist, but will at least eliminate bug that motivated #77 for me.

pawamoy:
Hey @tlambert03, I'm back :slightly_smiling_face: Busy weeks, sorry.

So, I wanted to try another approach first, and I did, but eventually I like what you did here. I still mixed a bit of both approach, I'll push directly on this PR so you can see, if that's OK with you!

tlambert03:
> Hey @tlambert03, I'm back 🙂 Busy weeks, sorry.

no worries, welcome back :)

> I still mixed a bit of both approach, I'll push directly on this PR so you can see, if that's OK with you!

awesome.  yes push away!

pawamoy:
So! I've pushed my changes.

Here's an explanation:

- merged "set members" mixins to bring the same functionality for both modules and collections, without spaghetti code
- in the finder, `find_package` does all the work of returning the right wrapper `Package` or `NamespacePackage`
- `Package` gained a `stubs` attribute
- the two previous points allow us to handle top modules *along with their stubs, even when `submodules=False`*. I've previously tried to do it differently but it didn't take `submodules` into account (btw I'm starting to doubt the usefulness of `submodules=False`...)
- additional test for modules with stubs directly in a search path

pawamoy (approved):
Let me know what you think! Maybe we could add comments to document the parts that are difficult to understand?

src/griffe/mixins.py
```diff
@@ -70,34 +70,15 @@ def __setitem__(self, key: str | Sequence[str], value) -> None:  # noqa: WPS231
...
-            name = parts[0]
-            if name in self.members:  # type: ignore[attr-defined]
-                for alias in self.members[name].aliases.values():  # type: ignore[attr-defined]
-                    alias.target = value
-            self.members[name] = value  # type: ignore[attr-defined]
-            value._modules_collection = self  # noqa: WPS437
+            if self.__class__.__name__ == "ModulesCollection":
+                value._modules_collection = self  # noqa: WPS437
+            else:
+                value.parent = self
```
pawamoy, line 81: Maybe we can set an attribute on the classes, for a more performant check. Let me do that.

tlambert03 (commented):
Yeah, looks great!   I like the changes.  This is good by me 👍

pawamoy:
Great, thank you! I'll merge and cut a new release :slightly_smiling_face:

Issue #84, opened 2022-06-09
Pth file formatting breaks griffe


samrensenhouse:
**Describe the bug**
This pth file https://github.com/mhammond/pywin32/blob/main/pywin32.pth seems to break `_handle_pth_file`. It errors out when checking if the directory exists.

```
  File "...\__pypackages__\3.8\lib\mkdocstrings_handlers\python\handler.py", line 182, in collect
    loader = GriffeLoader(
  File "...\__pypackages__\3.8\lib\griffe\loader.py", line 95, in __init__
    self.finder: ModuleFinder = ModuleFinder(search_paths)
  File "...\__pypackages__\3.8\lib\griffe\finder.py", line 59, in __init__
    self._extend_from_pth_files()
  File "...\__pypackages__\3.8\lib\griffe\finder.py", line 239, in _extend_from_pth_files
    self._append_search_path(_handle_pth_file(item))
  File "...\__pypackages__\3.8\lib\griffe\finder.py", line 282, in _handle_pth_file
    if added_dir.exists():
  File "...\AppData\Local\Programs\Python\Python38\lib\pathlib.py", line 1383, in exists
    self.stat()
  File "...\AppData\Local\Programs\Python\Python38\lib\pathlib.py", line 1189, in stat
    return self._accessor.stat(self)
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: "# .pth file for the PyWin32 extensions\nwin32\nwin32\\lib\nPythonwin\n# And some hackery to deal with environments where the post_install script\n# isn't run.\nimport pywin32_bootstrap"

```

**To Reproduce**
1. Install pywin32 package
2. Run griffe through mkdocstrings

mkdocs.yml
```
site_name: "Example"

nav:
- Home: index.md

plugins:
- mkdocstrings:
    default_handler: python

```

index.md
```
::: src.some_file.SomeClass
```

**Expected behavior**
Regardless of whether pywin32 is installed I would expect griffe to run successfully. If I uninstall the package, it works successfully.

**System (please complete the following information):**
- `griffe` version: 0.20.0
- Python version: 3.8
- OS: Windows

pawamoy:
Thanks for the report! I see there's already a PR, will review it now :slightly_smiling_face:

sanand0:
I'm facing this issue at the moment. Thrilled to see this fix! I'd be grateful for a release on pypi 🙏

PR #85, opened 2022-06-09
fix: ignore comments in .pth files

GabDug:
Fixes https://github.com/mkdocstrings/griffe/issues/84

Had the same issue with pdbpp on OS X, this made the trick.

This is a very naive implementation for single-line comments, feedback welcome.

### DIFF
--- a/src/griffe/finder.py
+++ b/src/griffe/finder.py
@@ -286,7 +286,7 @@ def _module_depth(name_parts_and_path: NamePartsAndPathType) -> int:
 
 def _handle_pth_file(path):
     # support for .pth files pointing to a directory
-    instructions = path.read_text().strip("\n").split(";")
+    instructions = list(filter(lambda l: not l.strip().startswith("#"), path.read_text().strip("\n").split(";")))
     added_dir = Path(instructions[0])
     if added_dir.exists():
         return added_dir


pawamoy:
Thanks for the PR!

So indeed the initial implementation was very naive. I've looked it up and found [these docs](https://docs.python.org/3/library/site.html):

> A path configuration file is a file whose name has the form name.pth and exists in one of the four directories mentioned above; its contents are additional items (one per line) to be added to sys.path. Non-existing items are never added to sys.path, and no check is made that the item refers to a directory rather than a file. No item is added to sys.path more than once. Blank lines and lines beginning with # are skipped. Lines starting with import (followed by space or tab) are executed.

I'll suggest some code modifications a bit later, unless you want to tackle it and follow the spec (not hard anyway).

src/griffe/finder.py
```diff
@@ -284,12 +286,25 @@ def _module_depth(name_parts_and_path: NamePartsAndPathType) -> int:
...
+    # Lines starting with import (followed by space or tab) are executed.
+    directories = []
+    for line in path.read_text().strip().splitlines(keepends=False):
+        line = line.strip()
+        if line and not line.startswith("#") and not _re_import_line.search(line):
+            if os.path.exists(line):
+                directories.append(Path(line))
+    if directories:
+        return directories
     raise UnhandledPthFileError(path)
```
pawamoy, line 308: Hmm, maybe we shouldn't raise at all. If no directories are found, just let it flow :thinking: ?

GabDug: Hello, thanks and sorry for the late response!
I do agree, we should probably let it flow, as users have little control over dependencies `.pth` files.

pawamoy: Done.

Issue #92, opened 2022-07-07
Module Exports are None


DarkaMaul:
**Describe the bug**

When using `mkdocstrings-python`, the generation of the documentation fails because of a griffe error.

**To Reproduce**

1. Create the following layout:
```
reproduce/
__init__.py
bar.py
```
2. In `__init__.py`

```python
from reproduce.bar import Foo

__all__ = [
    Foo,
]
```

3. In `bar.py`
```python
class Bar:
    pass
```

4. Run `main.py`:

```python
from griffe.loader import GriffeLoader
loader = GriffeLoader(search_paths=["/tmp/griffe/reproduce/"])
m = loader.load_module("reproduce")
loader.expand_exports(m)
```
5. See error

```console
File ~/.virtualenvs/tmp-156b6e7441e38e4/lib/python3.10/site-packages/griffe/loader.py:210, in GriffeLoader.expand_exports(self, module)
    208     next_module = self.modules_collection[module_path]
    209     self.expand_exports(next_module)
--> 210     expanded |= next_module.exports
    211 else:
    212     expanded.add(export)

TypeError: unsupported operand type(s) for |=: 'set' and 'NoneType'

```

**Expected behavior**

No error. ;)
**System (please complete the following information):**
- `griffe` version: 0.22 (master)
- Python version: 3.9 / 3.10
- OS: Linux

** Potential Fix**

I don't know why but the exports of the module are set as None.
A potential fix is to check that the value is not None before trying to perform the set operation.

```diff
diff --git a/src/griffe/loader.py b/src/griffe/loader.py
index 7095409..8ac6f15 100644
--- a/src/griffe/loader.py
+++ b/src/griffe/loader.py
@@ -207,7 +207,8 @@ class GriffeLoader:
                 module_path = export.full.rsplit(".", 1)[0]  # remove trailing .__all__
                 next_module = self.modules_collection[module_path]
                 self.expand_exports(next_module)
-                expanded |= next_module.exports
+                if next_module.exports is not None:
+                    expanded |= next_module.exports
             else:
                 expanded.add(export)
         module.exports = expanded
```

pawamoy:
Hi, thanks for the report.

The `__all__` variable is supposed to contain strings, not objects directly:

```python
from reproduce.bar import Foo

__all__ = [
    "Foo",
]
```

Could you try this and tell us if the issue still occur?

DarkaMaul:
Yep, the behavior is the same

[...]

PR #101, opened 2022-09-16
fix: expand empty module failed when `next_module.exports` is `None`

hfudev:


pawamoy:
Hi, thanks a lot!

For reference, this was reported in #92. Could you show us what code triggers the error for you? Thanks

hfudev:
@pawamoy Thanks! you're right. I'm using the `__all__` with the actual object, not the string. After switched to string the documentation is generated automatically.

pawamoy:
Great, thank you for reporting back :slightly_smiling_face:

Issue #75, opened 2022-05-29
Feature: API diffing utils & change detection.


tlambert03:
Continuing conversation started [here](https://github.com/mkdocstrings/griffe/issues/73#issuecomment-1140473155) ...

One of the things I'm most interested in using griffe for is detecting API changes.  This is one of the main things mentioned in the [README Todo](https://github.com/mkdocstrings/griffe#todo), so is clearly a major interest to @pawamoy as well.  There's a lot of subtleties to discuss here, so just starting this issue as courtesy/placeholder for this topic.

I've started an experimental repo to play with this here: https://github.com/tlambert03/griffediff ... but would be happy to port that work into this repo later if it's decided that that is the best place for it.  (or, alternatively, it could also be an independent repo, but still live in the mkdocstrings org).  There's lots of discussion to be had about approach there too.

I'll also mention https://github.com/Carreau/frappuccino.  I know that @carreau there wanted to do something purely static/AST-based (which is what griffe would be doing particularly well here), but eventually went to dynamic inspection due to difficulties in _actually_ being able to detect API changes statically.  (for example, a simple decorator can completely change an AST node at runtime... so the actual scope/usefulness of static API detection is definitely an open question)

pawamoy:
It's true that Python is (very) dynamic, and sometimes you have no choice but to actually *execute* things to get their actual representation, the one that users will see and use. But I think that a lot of projects can still be perfectly scanned with static analysis only. I myself very rarely use complex inheritance, decorations, re-binding/wrapping/re-assignments, etc. Most of the time my code is vanilla code. In some cases, visiting the AST is even *necessary*, for example to be able to pick up `@overload`-decorated methods (multiple signatures), which are simply overwritten when executing code. In any case, Griffe is able to both visit ASTs *and* introspect code (importing it), so I think it gives a solid foundation for API-diffing.

pawamoy:
I've read your code @tlambert03 and I like the way you use git to provide a reusable way to diff between two commits. This is very library-friendly and extendable.

tlambert03:
cool.  Yeah, I'm hoping that will help in a CI context too.  (i.e. run a github action and check the working tree against the main branch, and if a breaking change is detected, make sure there's a "`!`" in the PR title, or some other conventional commit marker of breaking change)

I do _also_ think that it would be nice to compare/cache json artifacts, as you discuss in the README.

pawamoy:
Hi @tlambert03, I hope you're doing well!

[...]

PR #105, opened 2022-10-02
feat: Add function to find API breaking changes

pawamoy:
Co-authored-by: Talley Lambert <talley.lambert@gmail.com>

---

@tlambert03, following discussion in #75, I've taken your code in griffediff, updated it to my taste (I hope you don't mind :sweat:) and credited you of course. This PR is still a draft anyway, just experimenting a bit.

What I envision is:

- utility to yield more than breakages: yield everything that changed or was added/removed (for auto-documentation purposes) and attach these data to objects
- also attach detected breakages to objects
- utility to yield forward-compatibility hints: "set these parameters as keyword-only" here and there, etc., and attach these data to objects. These hints could be displayed in docs auto-generated through mkdocstrings when serving only, to inform the developer :thinking:
- of course: add CLI options to show hints and breakages

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,25 +70,25 @@ See [the Usage section](https://mkdocstrings.github.io/griffe/usage/#with-python
     - Flat JSON
     - JSON schema
 - API diff:
-    - Mecanism to cache APIs? Should users version them, or store them somewhere (docs)?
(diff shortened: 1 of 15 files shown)


tlambert03:
nice!  looking good! :)

PR #126, opened 2023-01-17
fix: support editable install with multiple roots

gilfree:
Python packages may have multiple root modules where each is installed to its own folder under `site-packages`.

When using new setuptools editable install in such cases griffe, when used by mkdocstrings fails to find the correct paths.

The root cause it that griffe assumed a variable named `MAPPING` in the `.py` file created by the editable install, and assumed that this variable is a dict with single entry.

When a package has multiple roots - then this dictionary contain multiple entries.

This PR aims to handle such cases.

I have no knowledge about `editables` so I could not fix that case.

The unit test added simulates the case, and this has been tested against a real complex package with multiple roots using setuptools 65 editable install.

### DIFF
--- a/src/griffe/finder.py
+++ b/src/griffe/finder.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import ast
 import os
 import re
 import sys
@@ -264,12 +265,13 @@ class ModuleFinder:
                     for directory in _handle_pth_file(item):
                         self._append_search_path(directory)
 
-    def _extend_from_editable_modules(self):
+    def _extend_from_editable_modules(self):  # noqa: WPS231
         for path in self.search_paths:  # noqa: WPS440
             for item in self._contents(path):
                 if item.stem.startswith(("__editables_", "__editable__")) and item.suffix == ".py":
                     with suppress(UnhandledEditableModuleError):
-                        self._append_search_path(_handle_editable_module(item))
+                        for editable_path in _handle_editable_module(item):
+                            self._append_search_path(editable_path)
 
     def _filter_py_modules(self, path: Path) -> Iterator[Path]:
         for root, dirs, files in os.walk(path, topdown=True):
@@ -299,7 +301,6 @@ class ModuleFinder:
 _re_pkgresources = re.compile(r"(?:__import__\([\"']pkg_resources[\"']\).declare_namespace\(__name__\))")
 _re_pkgutil = re.compile(r"(?:__path__ = __import__\([\"']pkgutil[\"']\).extend_path\(__path__, __name__\))")
 _re_import_line = re.compile(r"^import[ \t]")
-_re_mapping_line = re.compile(r"^MAPPING = \{['\"].+['\"]: +['\"](.+)['\"]\}")
 
 
 # TODO: for better robustness, we should load and minify the AST
@@ -338,22 +339,24 @@ def _handle_editable_module(path: Path):  # noqa: WPS231
         editable_lines = path.read_text(encoding="utf8").splitlines(keepends=False)
     except FileNotFoundError:
         raise UnhandledEditableModuleError(path)
+
     if path.name.startswith("__editables_"):
         # support for how 'editables' writes these files:
         # example line: F.map_module('griffe', '/media/data/dev/griffe/src/griffe/__init__.py')
         new_path = Path(editable_lines[-1].split("'")[3])
         if new_path.exists():  # TODO: could remove existence check
             if new_path.name.startswith("__init__"):
-                return new_path.parent.parent
-            return new_path
+                return [new_path.parent.parent]
+            return [new_path]
     elif path.name.startswith("__editable__"):
         # support for how 'setuptools' writes these files:
-        # example line: MAPPING = {'griffe': '/media/data/dev/griffe/src/griffe'}
-        for line in editable_lines:
-            match = _re_mapping_line.match(line)
-            if match:
-                new_path = Path(match.group(1))
-                if new_path.exists():  # TODO: could remove existence check
-                    return new_path.parent
-                break
+        # example line: MAPPING = {'griffe': '/media/data/dev/griffe/src/griffe', 'briffe': '/media/data/dev/griffe/src/briffe'}
+        parsed_module = ast.parse(path.read_text())
+        for node in parsed_module.body:
+            if isinstance(node, ast.Assign):
+                if isinstance(node.targets[0], ast.Name) and node.targets[0].id == "MAPPING":
+                    if isinstance(node.value, ast.Dict):
+                        return [
+                            Path(constant.s).parent for constant in node.value.values if isinstance(constant, ast.Str)
+                        ]
     raise UnhandledEditableModuleError(path)
--- a/src/griffe/finder.py
+++ b/src/griffe/finder.py
@@ -339,7 +339,6 @@ def _handle_editable_module(path: Path):  # noqa: WPS231
         editable_lines = path.read_text(encoding="utf8").splitlines(keepends=False)
     except FileNotFoundError:
         raise UnhandledEditableModuleError(path)
-
     if path.name.startswith("__editables_"):
         # support for how 'editables' writes these files:
         # example line: F.map_module('griffe', '/media/data/dev/griffe/src/griffe/__init__.py')
--- a/tests/test_finder.py
+++ b/tests/test_finder.py
@@ -94,7 +94,7 @@ def test_editables_file_handling(tmp_path):
     """
     pth_file = tmp_path / "__editables_whatever.py"
     pth_file.write_text("hello\nF.map_module('griffe', 'src/griffe/__init__.py')")
-    assert _handle_editable_module(pth_file) == Path("src")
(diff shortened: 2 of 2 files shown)


gilfree:
Hi. 😄 

My first attempt in PR here. 
I hope I was able to properly follow the contributing guide. 
If there is something that needs to be changed - let me know.

src/griffe/finder.py
```diff
@@ -268,8 +269,12 @@ def _extend_from_editable_modules(self):
...
             for item in self._contents(path):
                 if item.stem.startswith(("__editables_", "__editable__")) and item.suffix == ".py":
-                    with suppress(UnhandledEditableModuleError):
-                        self._append_search_path(_handle_editable_module(item))
+                    self._extend_from_an_editable_module(item)
+
+    def _extend_from_an_editable_module(self, item: Path):
+        with suppress(UnhandledEditableModuleError):
+            for editable_path in _handle_editable_module(item):
+                self._append_search_path(editable_path)
```
pawamoy, line 277: Since the function is not reused elsewhere, no need to split it out.

src/griffe/finder.py
```diff
@@ -338,22 +343,21 @@ def _handle_editable_module(path: Path):  # noqa: WPS231
         editable_lines = path.read_text(encoding="utf8").splitlines(keepends=False)
     except FileNotFoundError:
         raise UnhandledEditableModuleError(path)
+
```
pawamoy, line 346: Please remove this style changed :smile: 
We should either space the code in the whole function, or not at all :stuck_out_tongue:

tests/test_finder.py
```diff
@@ -94,7 +94,7 @@ def test_editables_file_handling(tmp_path):
     """
     pth_file = tmp_path / "__editables_whatever.py"
     pth_file.write_text("hello\nF.map_module('griffe', 'src/griffe/__init__.py')")
-    assert _handle_editable_module(pth_file) == Path("src")
+    assert list(_handle_editable_module(pth_file)) == [Path("src")]
```
pawamoy, line 97: `_handle_editable_module` now returns a list, right? If true, then I don't think we need to wrap it in `list` again?

tests/test_finder.py
```diff
@@ -105,4 +105,15 @@ def test_setuptools_file_handling(tmp_path):
     """
     pth_file = tmp_path / "__editable__whatever.py"
     pth_file.write_text("hello\nMAPPING = {'griffe': 'src/griffe'}")
-    assert _handle_editable_module(pth_file) == Path("src")
+    assert list(_handle_editable_module(pth_file)) == [Path("src")]
```
pawamoy, line 108: Same remark here about wrapping in `list`.

tests/test_finder.py
```diff
@@ -105,4 +105,15 @@ def test_setuptools_file_handling(tmp_path):
...
+
+
+def test_setuptools_file_handling_multiple_paths(tmp_path):
+    """Assert editable modules by `setuptools` are handled when multiple packages are installed in the same editable.
+
+    Parameters:
+        tmp_path: Pytest fixture.
+    """
+    pth_file = tmp_path / "__editable__whatever.py"
+    pth_file.write_text("hello\nMAPPING = {'griffe': 'src1/griffe', 'briffe':'src2/briffe'}")
```
pawamoy, line 118: If we go my AST parsing suggestion above, we should add real statements before and after the `MAPPING` assignment, and write the assignment on multiple lines, something like:

```suggestion
    pth_file.write_text("print('hello')\nMAPPING = {\n'griffe': 'src1/griffe',\n'briffe':'src2/briffe'\n}\nprint('world')\n")
```

tests/test_finder.py
```diff
@@ -105,4 +105,15 @@ def test_setuptools_file_handling(tmp_path):
...
+    assert list(_handle_editable_module(pth_file)) == [Path("src1"), Path("src2")]
```
pawamoy, line 119: Same remark here about wrapping in `list`.

src/griffe/finder.py
```diff
@@ -338,22 +343,21 @@ def _handle_editable_module(path: Path):  # noqa: WPS231
...
-            match = _re_mapping_line.match(line)
-            if match:
-                new_path = Path(match.group(1))
-                if new_path.exists():  # TODO: could remove existence check
-                    return new_path.parent
-                break
+            if line.startswith("MAPPING"):
+                paths = ast.literal_eval(line.split(" = ", 1)[1])
+                assert isinstance(paths, dict)
+                return [Path(pt).parent for pt in paths.values()]
```
pawamoy, line 362: At this point I feel like we should parse the entire file with AST (without any evaluation), and iterate on the module's body to find the `MAPPING` assignment node. Who knows if setuptools won't split the assignment on multiple lines in some cases, or in the future?

It would look like this:

```python
parsed_module = ast.parse(path.read_text())
for node in parsed_module.body:
    if isinstance(node, ast.Assign) and node.targets[0].id == "MAPPING":
        return [Path(constant.value).parent for constant in node.value.values]
```

pawamoy (changes requested):
Thanks, that's great! You did very well for your first contribution here :rocket: 

I have a few suggestions below :slightly_smiling_face:

src/griffe/finder.py
gilfree: Ok, The code complexity checking claims nesting these loops creates a too complex function. Should I silent that warning?

src/griffe/finder.py
gilfree: I have changed to full AST parsing, I agree it will be more stable.

A concern I do have at this point is that the change by setuptools or editables or any other package will be bigger than that. 

It is probably not for this MR, and I don't understand the ModuleFinder mechanism enough, but I think it might probably  be augmented by using the sys.meta_hook, and in that case there is no need to specially handle each editable install variant. 

If you prefer, I *think* the code below may replace the `_extend_from_editable_modules`

```python
    def find_package(self, module_name: str) -> Package | NamespacePackage:  # noqa: WPS231
        """Find a package or namespace package.

        Parameters:
            module_name: The module name.

        Raises:
            ModuleNotFoundError: When the module cannot be found.

        Returns:
            A package or namespace package wrapper.
        """
        filepaths = [
            Path(module_name),
            # TODO: handle .py[cod] and .so files?
            Path(f"{module_name}.py"),
        ]
       # Add meta path
        specs_paths = []
        for finder in sys.meta_path:
            spec = finder.find_spec(module_name,None)
            if spec and spec.submodule_search_locations:
                specs_paths.extend(Path(path).parent for path in spec.submodule_search_locations)
        namespace_dirs = []
        for path in self.search_paths + specs_paths:  # noqa: WPS440
            path_contents = self._contents(path)
            if path_contents:
                for choice in filepaths:
                    abs_path = path / choice
                    if abs_path in path_contents:
                        if abs_path.suffix:
                            stubs = abs_path.with_suffix(".pyi")
                            return Package(module_name, abs_path, stubs if stubs.exists() else None)
                        else:
                            init_module = abs_path / "__init__.py"
                            if init_module.exists() and not _is_pkg_style_namespace(init_module):
                                stubs = init_module.with_suffix(".pyi")
                                return Package(module_name, init_module, stubs if stubs.exists() else None)
                            namespace_dirs.append(abs_path)

        if namespace_dirs:
            return NamespacePackage(module_name, namespace_dirs)

        raise ModuleNotFoundError(module_name)
```

tests/test_finder.py
gilfree: Fixed.

tests/test_finder.py
gilfree: Changed.

tests/test_finder.py
gilfree: Fixed

gilfree (commented):
I have tried to fix according to your comments.
Thanks.

src/griffe/finder.py
pawamoy: Ha, I was wondering if you split it for that reason! Yes, just `# noqa: WPS231` it please :slightly_smiling_face:

src/griffe/finder.py
```diff
@@ -338,22 +339,24 @@ def _handle_editable_module(path: Path):  # noqa: WPS231
         editable_lines = path.read_text(encoding="utf8").splitlines(keepends=False)
     except FileNotFoundError:
         raise UnhandledEditableModuleError(path)
+
```
pawamoy, line 342:
```suggestion
```

src/griffe/finder.py
pawamoy: I went ahead and committed a suggestion directly from GitHub :slightly_smiling_face:

pawamoy (approved):
OK, LGTM, thanks so much :rocket:

src/griffe/finder.py
gilfree: This is a small change on the griffe code - just added the lines below (which I wrote myself) to the function from griffe, and iterated also on `specs_paths`

```pypthon
        specs_paths = []
        for finder in sys.meta_path:
            spec = finder.find_spec(module_name,None)
            if spec and spec.submodule_search_locations:
                specs_paths.extend(Path(path).parent for path in spec.submodule_search_locations)
```

PR #127, opened 2023-01-23
docs: minor typo fixes

thatlittleboy:
Just a couple of typo fixes to the docs.

And I realized the output of `griffe dump` is an object, not an array anymore, so updated the docs to reflect that


pawamoy (approved):
Thanks a lot!

PR #135, opened 2023-03-02
feat: Forward class attribute docstrings to instances.

### LINKED ISSUE #128: Question: can I do anything about docstrings for attributes being picked up from last location
Hey there,

I have been using mkdocstrings with mkdocs and I've hit an interesting predicament.
When documenting attributes in Python code, griffe is able to pick up docstrings from within the class body and from within the dunder method `__init__`. I.e., in the code below, both `foo` and `bar` would be documented:

```py
class C:
    foo = 42
    """A number I enjoy."""

    def __init__(self):
        self.bar = 73
        """Another number I enjoy."""
```

However, in some cases I'll set a default value in the class body and then I'll do some initialisation inside `__init__` and, in that case, griffe will pick the docstring that is used last... Even if one of the cases _does not_ include a docstring.
E.g., in the situation below `foo` would not have a docstring:

```py
class C:
    foo = 42
    """A number I enjoy."""

    def __init__(self, number):
        self.foo = number
```

I would expect `foo` to be documented with the docstring “A number I enjoy.”.
However, if I move the docstring to inside `__init__` or the assignment to be _after_ the method `__init__`, then it'll work.
Is this a bug? Is this an intentional design decision? Can I do anything about it?

Thanks a lot for your time!

P.S. for reference, the original exploration that led to me opening this issue is https://github.com/Textualize/textual/issues/1572.

rodrigogiraoserrao:
This will fix #128.

Thanks for the guidance, @pawamoy.
I tried doing what was needed (follow commit message format conventions, add the test in the correct place, etc) but I might have missed something.
If so, please let me know.

### DIFF
--- a/src/griffe/agents/visitor.py
+++ b/src/griffe/agents/visitor.py
@@ -590,6 +590,12 @@ class Visitor(BaseVisitor):
                 if isinstance(node.parent, (ast.If, ast.ExceptHandler)):  # type: ignore[union-attr]
                     continue  # prefer "no-exception" case
 
+                parent.members[name].labels |= labels
+
+                # forward previous docstring instead of erasing it
+                if parent.members[name].docstring and not docstring:
+                    docstring = parent.members[name].docstring
+
             attribute = Attribute(
                 name=name,
                 value=value,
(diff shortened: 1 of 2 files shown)


pawamoy (approved):
That's great, thanks a lot! :rocket:

pawamoy:
I've merged as "fix" rather than "feat".

rodrigogiraoserrao:
> I've merged as "fix" rather than "feat".

Of course, sorry for getting that wrong.

pawamoy:
> Of course, sorry for getting that wrong.

Oh that's more a matter of opinion/maintenance rather than correctness, no worries :+1:
"Fix" conveys that this is what we support, and not something else (previous behavior for example).

PR #140, opened 2023-03-31
feat: Implement basic handling of Alias for breaking changes

Georggi:


pawamoy:
Thanks a lot! Could you add some tests?

pawamoy:
Thanks for the update. Since both tests are very similar, I'd prefer that we combine them, if possible (easier to maintain). I'll try to do that if you don't mind.

pawamoy:
Just pushed my changes. Let me know what you think. IMO it's looking good and we can merge :slightly_smiling_face:

Georggi:
My only concern is that we removed a more basic case with no alias. I think there should still be a separate assert at least that would not include init module.

pawamoy:
Good point, I've added back tests without aliases :slightly_smiling_face:

pawamoy:
Thanks again!

PR #149, opened 2023-04-16
fix: Allow Raises and Warns sections to start with a newline

viccie30:
According to the documentation it's possible to start a description with a newline.
However, this did not work properly for Raises and Warns sections, because of a stray space in the regex to split the annotation and its description.

### DIFF
--- a/src/griffe/docstrings/google.py
+++ b/src/griffe/docstrings/google.py
@@ -310,7 +310,7 @@ def _read_raises_section(
     annotation: str | Name | Expression
     for line_number, exception_lines in block:
         try:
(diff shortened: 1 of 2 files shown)


pawamoy:
Nice catch, thanks a lot!

viccie30:
I just realised I missed a word in my commit message, so I've force pushed the same commit with the missing word in the message added.

pawamoy:
I've added a test for it. LGTM, WDYT?

viccie30:
Looks good to me as well!

pawamoy:
I've also found out warnings sections were never parsed. Just released both fixes as 0.27.1.

PR #150, opened 2023-04-17
Respect ClassVar variable annotation

viccie30:
According to [PEP 526](https://peps.python.org/pep-0526/#class-and-instance-variable-annotations) annotated attributes in a class body are instance variables by default, unless explicitly overridden by using [`typing.ClassVar`](https://docs.python.org/3/library/typing.html#typing.ClassVar).

This PR makes 2 changes:
1. Honor `typing.ClassVar` in class variable annotations.
2. Merge the annotations in the class body and in `__init__()`.

### DIFF
--- a/src/griffe/agents/visitor.py
+++ b/src/griffe/agents/visitor.py
@@ -42,13 +42,14 @@ from griffe.dataclasses import (
     Parameters,
 )
 from griffe.exceptions import AliasResolutionError, CyclicAliasError, LastNodeError, NameResolutionError
+from griffe.expressions import Expression
 from griffe.extensions import Extensions
 
 if TYPE_CHECKING:
     from pathlib import Path
 
     from griffe.docstrings.parsers import Parser
-    from griffe.expressions import Expression, Name
+    from griffe.expressions import Name
 
 
 builtin_decorators = {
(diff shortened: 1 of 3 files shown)


pawamoy:
Hey @viccie30, thanks for the PR!

I think we need to distinguish three cases:

- attribute is annotated with `ClassVar`: set `class-attribute` label only. If attribute is redfined in `__init__`, don't bother checking: that's a user error, keep the current flow (add `instance-attribute` label, etc.)
- attribute is only annotated: since the attribute is not accessible through the class at runtime, but only through instances, set `instance-attribute` label
- attribute is annotated and assigned: set `class-attribute` label, and keep the current flow (add `instance-attribute` label if defined again in `__init__` etc.)

viccie30:
Thanks for looking into the PR.

I agree that there's 3 cases. Case 1 is user error, like you say. Case 2 is already supported in this PR, through line 566 in src/griffe/agents/visitor.py.

I'm not sure what you mean with case 3. According to the PEP if a non-ClassVar annotated attribute at class scope gets a value assigned at class scope, it's still an instance variable and not a class variable. This PR handles it as such. I have added new test cases to check this.

viccie30:
I have added a branch in `handle_attribute()` to mark initialized annotated variables as class variables.

pawamoy:
I've added the `instance-attribute` label when the attribute has a value but is not annotated as `ClassVar`. LGTM now, merging, thanks!

Issue #163, opened 2023-05-23
Expression.full() errors if any Expressions inside it


machow:
**Describe the bug**

For griffe v0.28.1, when `Expression.full()` gets called, if it contains any expressions within it, then it attempts to call `.canonical` on them, which produces an AttributeError:

```
AttributeError: 'Expression' object has no attribute 'canonical'
```

Here's an example that I used to check against:

```python
a: Union[list[str], str]
```

pawamoy:
Damn, how am I not detecting this with the tests suite yet. Thanks for the report, will fix asap

machow:
No worries! I was able to fix pretty fast and submit a PR in a sec! (No worries if you end up fixing another way)

PR #164, opened 2023-05-23
fix: full method over expressions of expressions

machow:
This PR addresses #163, by calling full on any child expressions. Happy to make any tweaks :)


pawamoy:
Thanks a lot! I pushed minor changes to have CI pass. Will merge right after :slightly_smiling_face:

Issue #167, opened 2023-06-06
numpy parser parameter table breaks if there is an empty line between parameters


machow:
**Describe the bug**

From what I can tell, the numpydoc parser supports empty lines between parameter descriptions. e.g.

```python
"""
Parameters
------------
x: int
    a thing
                                         <--- note this empty line
y: float
    another thing
"""
```

However, with griffe's numpy parser this triggers the end of the parameters table.

**One possible solution**

If we instead use the following rule to indicate a new section, it should be able to keep the table intact, without changing the current behavior too much.

* detect directly the next section of documentation (e.g. line has words, and the next line starts with `---`)

**Breaking cases**

Note that with this solution, these kinds of docstrings from the unit tests could be parsed differently:

```
Returns
-------
only_item : type
    Description.



Something.
```

Currently, these get parsed as 2 sections (a Returns and a generic Text section).  However, AFAICT, the numpydoc parser just considers the "Something." to be part of the Returns section, since sections must be separated by headings.

```python
from numpydoc.docscrape import NumpyDocString

docstring = """
Returns
-------
only_item : type
    Description.



Something.
"""

doc = NumpyDocString(docstring)
list(doc.items())
```

WDYT? Happy to find the right balance of current behavior / supporting table description variations!

pawamoy:
Thanks for the report!

I find it weird that 

```
Returns
-------
only_item : type
    Description.



Something.
```

...would be parsed as a single Returns section. How do you intertwine prose and semantic sections? Is that not possible with Numpydoc's spec? I think supporting blank lines in-between items is very niche. IMO it costs nothing to just remove the blank line, and it even makes the docstring more readable.

*However* this is just my opinion, and I'm fine with anything that matches the spec: I'm not using this docstring style after all, so those who actually use it know better :slightly_smiling_face: Reading again the spec at https://numpydoc.readthedocs.io/en/latest/format.html, it seems indeed that prose is not expected in-between sections, and therefore headings (and dash lines) are the true sections delimiters, not blank lines.

machow:
Hmm... I dug a bit deeper, and it looks like sphinx does allow including extra narrative (to a degree).

It seems like the rule they use to end a section ([sphinx code here](https://github.com/sphinx-doc/sphinx/blob/d3c91f951255c6729a53e38c895ddc0af036b5b9/sphinx/ext/napoleon/docstring.py#L1223)) is...

* Encounter a new section, OR
* 2 empty lines in the table block

[...]

PR #168, opened 2023-06-06
Fix numpy block linebreaks

machow:
This PR addresses #167. I'm happy to tweak / look into supporting different cases. (I've xfailed the test of what would be a breaking change, and added a test of the new supported behavior.)

### DIFF
--- a/src/griffe/docstrings/numpy.py
+++ b/src/griffe/docstrings/numpy.py
@@ -105,7 +105,13 @@ def _is_dash_line(line: str) -> bool:
     return not _is_empty_line(line) and _is_empty_line(line.replace("-", ""))
 
 
-def _read_block_items(docstring: Docstring, *, offset: int, allow_block_breaks:bool, **options) -> tuple[list[list[str]], int]:
+def _read_block_items(
+    docstring: Docstring,
+    *,
+    offset: int,
+    allow_block_breaks:bool,
+    **options: Any,                      # noqa: ARG001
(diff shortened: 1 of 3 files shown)


src/griffe/docstrings/numpy.py
```diff
@@ -648,6 +665,7 @@ def parse(
     *,
     ignore_init_summary: bool = False,
     trim_doctest_flags: bool = True,
+    allow_block_breaks: bool = False,
```
pawamoy, line 668: Shouldn't it be true to be prevent a breaking change? By the way I'm confused by the parameter name and description. I keep going back to the logic above to understand what it does. I think naming it `allow_section_blank_line` would be better :thinking: If true, then one needs *two* blank lines to break a section. WDYT?

machow: Hey, sorry for the wait--I was out of town last week. Agreed the parameter name is confusing, lemme go ahead and switch it to `allow_section_blank_line` and fix the default.

pawamoy: No problem! Lets not be sorry for life things ^^ Thanks for the follow-up :)

pawamoy:
Ugh, I included tests for inheritance in the main branch, but the feature is not merged yet... let me fix CI.

pawamoy (approved):
LGTM

pawamoy:
Thanks @machow!

Issue #96, opened 2022-08-12
Docstrings are not inherited


davfsa:
**Describe the bug**
Docstrings are not inherited. Considering this is not any special case and it defined by the Python standard, I thought it would be appropriate to open the issue here, but the end goal is for mkdocstrings.

**To Reproduce**
```py
class A:
    def foo(self):
        """This is a docstring"""

class B(A):
    def foo(self):
        # Inherited docstring defined in A.foo
```

```py
>>> from griffe.loader import GriffeLoader
>>> loader = GriffeLoader()
>>> module = loader.load_module("module")
>>> print(module.classes["A"].functions["foo"].docstring)
This is a docstring
>>> print(module.classes["B"].functions["foo"].docstring)
None
```

**Expected behavior**
Docstrings are inherited.

**System:**
- `griffe` version: 0.22.0
- Python version: 3.10.4
- OS: Linux

**Additional context**
N/A

machow:
It seems like griffe's behavior matches python's behavior

```python
class A:
    def foo(self):
        """This is a docstring"""

class B(A):
    def foo(self):
        # Inherited docstring defined in A.foo

        ...

# returns NoneType
type(B.foo.__doc__)
```

davfsa:
> It seems like griffe's behavior matches python's behavior
> 
> ```python
> class A:
>     def foo(self):
>         """This is a docstring"""
> 
> class B(A):
>     def foo(self):
>         # Inherited docstring defined in A.foo
> 
>         ...
> 
> # returns NoneType
> type(B.foo.__doc__)
> ```

That is correct, but griffle seems to be more like `inspect`, which does take into account inherited docstrings when using `inspect.getdocstring()`

As I initially stated, the end goal is `mkdocstrings`, so I can open an issue there instead if needed.

cc @pawamoy

pawamoy:
Hello @davfsa, sorry for the late reply and thanks for the report.
Thanks @machow as well for helping out.

Currently Griffe does not support any kind of inheritance. Since different users wish for different things, there is some work to do to specify exactly what we want to provide, and how it should be configurable.

Just wanted to let you know that I don't forget about issues. My time is limited but I'll eventually get to it!

gab832:
Hi @pawamoy, I'm Gerard from Cloudblue. We are also interested on the inherited methods. How is it going? Is there any plan to do it? Thanks for all the work you are doing!

pawamoy:
[...]

PR #170, opened 2023-06-15
feat: Support inheritance

### LINKED ISSUE #138: Inherited menber
# Context

Hey, first thanks for the great tools!

I commented recently on this [issue](https://github.com/mkdocstrings/pytkdocs/issues/58) in pytkdocs.

My end goal is to make work the `inherited_members` of mkdocstring work.

I was about to create a PR for pytkdocs but I saw that the new python handler of mkdostring is using griffe. 

So I was wondering :

* is ìnherited_menbers` or something similar is implemented on griffe ?
* If no, I will try to draft a PR to fix it (as I really need to be able to show inherited method for my project), where should I start ?

Thanks in advance

### LINKED ISSUE #153: Make inheritance of members from the parent class configurable when inspecting
Hi,

I'm developing a Cython project and therefore parsing the source-code directly in order to generate documentation (which is what the `Visitor` does if I'm correct) is not possible.
Thus, the only way is via inspecting.

However, the problem I encountered is that when inspection is done, Members of the Parent are also shown in the documentation for the class.
For example, consider the following:

```python
class Test(dict):
    """Inherit from dict"""

    def __init__(self):
        pass

    def test_func(self):
        """test_func"""
        pass


class Blaa(Exception):
    """Test exception"""
    def __init__(self):
        pass
```

It will render like this:

image

As the screenshot shows, any public function of the parent is also included.

**Describe alternatives you've considered**
Not really any good alternative when it comes to Cython projects, because everything is compiled and inspecting is the only way.

I feel for Inspection, this should be configurable. Configurable because sometimes it might make sense to include the functions from the Parent class.
I found [this issue](https://github.com/mkdocstrings/mkdocstrings/issues/78) which essentially wants just that: inheriting from the parent (which is not yet possible when `visiting` the source-code directly instead of `inspecting`, since the `inherited_members` functionality is [not yet implemented in mkdocstrings](https://github.com/mkdocstrings/mkdocstrings/issues/78#issuecomment-1511013734) if I got that right)
So, for those that would want to inherit the members, forcing inspection instead of visiting the source-code would be an option (although source-code snippets wouldn't show up anymore)

However in other cases, for example when inheriting from something like `dict` it doesn't really make much sense in my view. The user is already notified that it inherits from `dict` (by showing `Bases: dict` at the top), and should therefore look at the documentation for the `dict` class instead.

Now what I don't know: When `inherited_members` is supported in `mkdocstrings`, and I would set it to `false`, would I get the desired functionality I am describing here? (for both inspection and visiting) i.e. public functions of the parents wouldn't show up?
If that would be the case, well then I guess its just a matter of time when it gets supported, and probably nothing must be changed in `griffe` directly.

I poked around a bit in the source code and found a way to disable the inheritance of members when inspecting, by adding another condition in the `_pick_members` function in `src/griffe/agents/nodes.py`:

```python
    def _pick_member(self, name: str, member: Any) -> bool:
        return (
            name not in self.exclude_specials
            and member is not type
            and member is not object
            and id(member) not in self._ids
            # Only include it if this is really a member of the current object
            and name in vars(self.obj)
        )
```

I figured that `self.obj` will always be of kind `module` or `class` when it collects the children of the node, and if the member is not directly a part of `vars(self.obj)`, then it is probably inherited.
So the difference is: `inspect.getmembers(self.obj)` always returns all the members including the inherited ones for a class, but `vars(self.obj)` does not include the inherited ones. So we can just check if the member is actually in `vars(self.obj)`.
If the check is added and I rebuild the documentation, then all the inherited members go away.

Not sure if this is the best way to go about it, but that was the solution I found (excluding the fact that `inherited_members` may already handle this anyway)

What do you think about this?

pawamoy:
Closes #138
Closes #153 
Unblocks work on #96, https://github.com/mkdocstrings/python/issues/40, https://github.com/mkdocstrings/python/issues/13
Related to https://github.com/mkdocstrings/mkdocstrings/issues/459, https://github.com/mkdocstrings/mkdocstrings/issues/157, https://github.com/mkdocstrings/mkdocstrings/discussions/536

Tagging people who are probably interested: @samsja, @tazend, @cpcloud, @famura, @machow, @gab832, @analog-cbarber, @jayqi, @chrieke, @tristanlatr
Feel free to tag others. Also, subscribe to this PR to get future notifications: I'll explain the changes in this PR later so that we can discuss them. Or unsubscribe to stop getting notifications :slightly_smiling_face: 

---

Explanation:

- the inspector only picks up non-inherited members
- the visitor does not change at all
- inherited members are fetched on demand with `Object.inherited_members`
- inherited members are only fetched for classes, not instances (attributes), as instances have a type that can be used to point to the corresponding class
- to fetch inherited members, we resolve base classes (get the actual `Class`es representing them) and use C3 linearization to build the dict of inherited members according to Python's MRO (method resolution order)
- if multiple packages are involved in the inheritance, only members from pre-loaded packages (available in each module's `modules_collection`) are fetched
- this gives control to users, allowing to choose how deep in the class inheritance to go ("inherit from pydantic classes, but not further") by pre-loading packages/modules of their choice

---

mkdocstrings and the Python handler:

- we will be able to add an option `inherit_members: true/false`, false by default
- when false, the `members` attribute of objects will be used to filter members
- when true, the `all_members` property of objects will be used to filter members

---

Todo:

- [x] **tests!!!** :smile: 
    - [x] generic tests
    - [x] MRO tests
    - [x] dynamic class base tests
- [x] fuzzing on stdlib
- [x] fuzzing on popular packages
- [x] docs :nerd_face:

### DIFF
--- a/src/griffe/agents/inspector.py
+++ b/src/griffe/agents/inspector.py
@@ -295,7 +295,11 @@ class Inspector(BaseInspector):
         Parameters:
             node: The node to inspect.
         """
-        bases = [base.__name__ for base in node.obj.__bases__ if base is not object]
+        bases = []
+        for base in node.obj.__bases__:
+            if base is object:
+                continue
+            bases.append(f"{base.__module__}.{base.__qualname__}")
 
         class_ = Class(
             name=node.name,
--- a/src/griffe/agents/inspector.py
+++ b/src/griffe/agents/inspector.py
@@ -266,7 +266,7 @@ class Inspector(BaseInspector):
                 else:
                     child_name = getattr(child.obj, "__name__", child.name)
                     target_path = f"{child_module_path}.{child_name}"
-                self.current[child.name] = Alias(child.name, target_path)
+                self.current.set_member(child.name, Alias(child.name, target_path))
             else:
                 self.inspect(child)
 
@@ -306,7 +306,7 @@ class Inspector(BaseInspector):
             docstring=self._get_docstring(node),
             bases=bases,
         )
-        self.current[node.name] = class_
+        self.current.set_member(node.name, class_)
         self.current = class_
         self.generic_inspect(node)
         self.current = self.current.parent  # type: ignore[assignment]
@@ -434,7 +434,7 @@ class Inspector(BaseInspector):
                 docstring=self._get_docstring(node),
             )
         obj.labels |= labels
-        self.current[node.name] = obj
+        self.current.set_member(node.name, obj)
 
     def inspect_attribute(self, node: ObjectNode) -> None:
         """Inspect an attribute.
@@ -481,7 +481,7 @@ class Inspector(BaseInspector):
             docstring=docstring,
         )
         attribute.labels |= labels
-        parent[node.name] = attribute
+        parent.set_member(node.name, attribute)
 
         if node.name == "__all__":
             parent.exports = set(node.obj)
--- a/src/griffe/agents/nodes.py
+++ b/src/griffe/agents/nodes.py
@@ -318,6 +318,13 @@ class ObjectNode:
         return f"ObjectNode(name={self.name!r})"
 
     @cached_property
+    def path(self) -> str:
+        """The object's (Python) path."""
+        if self.parent is None:
+            return self.name
+        return f"{self.parent.path}.{self.name}"
+
+    @cached_property
     def kind(self) -> ObjectKind:
         """Return the kind of this node.
 
@@ -508,6 +515,7 @@ class ObjectNode:
             and member is not type
             and member is not object
             and id(member) not in self._ids
+            and name in vars(self.obj)
         )
 
 
--- a/src/griffe/c3linear.py
+++ b/src/griffe/c3linear.py
@@ -1,0 +1,117 @@
+"""Compute method resolution order. Implements `Class.mro` attribute."""
+
+# Copyright (c) 2019 Vitaly R. Samigullin
+# Adapted from https://github.com/pilosus/c3linear
+# Adapted from https://github.com/tristanlatr/pydocspec
+
+from __future__ import annotations
+
+from itertools import islice
+from typing import Deque, TypeVar
+
+T = TypeVar("T")
+
+
+class _Dependency(Deque[T]):
+    """A class representing a (doubly-ended) queue of items."""
+
+    @property
+    def head(self) -> T | None:
+        """Head of the dependency."""
+        try:
+            return self[0]
+        except IndexError:
+            return None
+
+    @property
+    def tail(self) -> islice:
+        """Tail od the dependency.
+
+        The `islice` object is sufficient for iteration or testing membership (`in`).
+        """
+        try:
+            return islice(self, 1, self.__len__())
+        except (ValueError, IndexError):
+            return islice([], 0, 0)
+
+
+class _DependencyList:
+    """A class representing a list of linearizations (dependencies).
+
+    The last element of DependencyList is a list of parents.
+    It's needed  to the merge process preserves the local
+    precedence order of direct parent classes.
+    """
+
+    def __init__(self, *lists: list[T | None]) -> None:
+        """Initialize the list.
+
+        Parameters:
+            *lists: Lists of items.
+        """
+        self._lists = [_Dependency(lst) for lst in lists]
+
+    def __contains__(self, item: T) -> bool:
+        """Return True if any linearization's tail contains an item."""
+        return any(item in lst.tail for lst in self._lists)
+
+    def __len__(self) -> int:
+        size = len(self._lists)
+        return (size - 1) if size else 0
+
+    def __repr__(self) -> str:
+        return self._lists.__repr__()
+
+    @property
+    def heads(self) -> list[T | None]:
+        """Return the heads."""
+        return [lst.head for lst in self._lists]
+
+    @property
+    def tails(self) -> _DependencyList:
+        """Return self so that `__contains__` could be called."""
+        return self
+
+    @property
+    def exhausted(self) -> bool:
+        """True if all elements of the lists are exhausted."""
+        return all(len(x) == 0 for x in self._lists)
+
+    def remove(self, item: T | None) -> None:
+        """Remove an item from the lists.
+
+        Once an item removed from heads, the leftmost elements of the tails
+        get promoted to become the new heads.
+        """
+        for i in self._lists:
+            if i and i.head == item:
+                i.popleft()
+
+
+def c3linear_merge(*lists: list[T]) -> list[T]:
+    """Merge lists of lists in the order defined by the C3Linear algorithm.
+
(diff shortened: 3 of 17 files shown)


src/griffe/agents/nodes.py
```diff
@@ -508,6 +508,7 @@ def _pick_member(self, name: str, member: Any) -> bool:
             and member is not type
             and member is not object
             and id(member) not in self._ids
+            and name in vars(self.obj)
```
pawamoy, line 511: From https://github.com/mkdocstrings/griffe/issues/153, thanks @tazend! The goal here is to only get members that are not inherited, and rely on the abstract (as in, visitor-inspector agnostic) `Object.inherited_members` property.

src/griffe/agents/inspector.py
```diff
@@ -287,7 +287,11 @@ def inspect_class(self, node: ObjectNode) -> None:
         Parameters:
             node: The node to inspect.
         """
-        bases = [base.__name__ for base in node.obj.__bases__ if base is not object]
+        bases = []
+        for base in node.obj.__bases__:
+            if base is object:
+                continue
+            bases.append(f"{base.__module__}.{base.__name__}")
```
pawamoy, line 294: We need the fully qualified name to resolve the bases later, in `Class.resolved_bases`.

src/griffe/c3linear.py
```diff
@@ -0,0 +1,118 @@
+"""Compute method resolution order. Implements `Class.mro` attribute."""
+
+# MIT License
+# Copyright (c) 2019 Vitaly R. Samigullin
+# Adapted from https://github.com/pilosus/c3linear
+# Adapted from https://github.com/tristanlatr/pydocspec
```
pawamoy, line 6: Thanks @tristanlatr :heart:

src/griffe/dataclasses.py
```diff
@@ -428,6 +432,22 @@ def is_kind(self, kind: str | Kind | set[str | Kind]) -> bool:
             kind = Kind(kind)
         return self.kind is kind
 
+    @cached_property
+    def inherited_members(self) -> dict[str, Alias]:
+        """Members that are inherited from base classes."""
+        if not isinstance(self, Class):
+            return {}
```
pawamoy, line 439: Only classes have "inherited" members. Inherited members won't be populated for class instances (attributes). Attributes have a type that points to the relevant class.

This is up to discussion, especially regarding inspection with dynamic base classes.

src/griffe/dataclasses.py
```diff
@@ -428,6 +432,22 @@ def is_kind(self, kind: str | Kind | set[str | Kind]) -> bool:
...
+        inherited_members = {}
+        for base in reversed(self.mro()):
+            for name, member in base.members.items():
+                inherited_members[name] = Alias(name, member, parent=self)  # type: ignore[arg-type]
```
pawamoy, line 443: Inherited members are stored (cached property) as aliases. This is important for back-pointers. We cannot store the targets (actual members) into the subclasses' members, because that would overwrite their parent attribute, making the object tree incorrect.

src/griffe/dataclasses.py
```diff
@@ -428,6 +432,22 @@ def is_kind(self, kind: str | Kind | set[str | Kind]) -> bool:
...
         return self.kind is kind
 
+    @cached_property
+    def inherited_members(self) -> dict[str, Alias]:
+        """Members that are inherited from base classes."""
+        if not isinstance(self, Class):
+            return {}
+        inherited_members = {}
+        for base in reversed(self.mro()):
+            for name, member in base.members.items():
```
pawamoy, line 442: I was wondering if we should iterate on *inherited* members instead of regular ones, as to recursively inflate inherited members of all classes in the chain, but that does not seem correct based on Python's MRO (C3 linearization), and that would mean a lot of back and forth (potentially exponential) between child and parent classes.

src/griffe/dataclasses.py
```diff
@@ -428,6 +432,22 @@ def is_kind(self, kind: str | Kind | set[str | Kind]) -> bool:
...
+                inherited_members[name] = Alias(name, member, parent=self)  # type: ignore[arg-type]
+        return inherited_members
+
+    @property
+    def all_members(self) -> dict[str, Object | Alias]:
```
pawamoy, line 447: I considered changing `members` to a property returning all members, and adding a `declared_members` property, but that would be a breaking change.

src/griffe/dataclasses.py
```diff
@@ -1392,6 +1415,29 @@ def parameters(self) -> Parameters:
...
+                resolved_base = self.modules_collection[base_path]
+                if resolved_base.is_alias:
+                    resolved_base = resolved_base.final_target
+            except (AliasResolutionError, CyclicAliasError, KeyError):
+                logger.debug(f"Base class {base_path} is not loaded, or not static, it cannot be resolved")
+            else:
+                resolved_bases.append(resolved_base)
+        return resolved_bases
+
+    def mro(self) -> list[Class]:
```
pawamoy, line 1434: We could consider caching the result. We do that a lot throughout the code base though, and already consume a lot of memory. We'll need to optimize at some point.

src/griffe/agents/inspector.py
pawamoy: You're right, good catch. I think I can re-use the object node (`node` here) to build the path of the object, since every node has a name and a pointer to its parent.

tests/test_inheritance.py
```diff
@@ -0,0 +1,80 @@
+"""Tests for class inheritance."""
```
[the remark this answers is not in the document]
pawamoy: Thank you so much! Yeah I was planning on adding more tests :slightly_smiling_face:

src/griffe/agents/inspector.py
pawamoy: Solved this by using the class' `__qualname__` attribute instead.

machow:
This looks great! One quick thought on...

> if multiple packages are involved in the inheritance, only members from pre-loaded packages

I wonder if it will be helpful downstream in the python handler to emit a warning, if someone uses `inherit_members: true` and is missing some pre-loaded packages. It makes sense to me that people would need to pre-load packages, though!

machow:
Ah, yeah that seems great! I think we might be thinking of the same thing, but using upstream and downstream in different ways 😅.

I was thinking it might be helpful to give a more specific message in mkdocs-python, like "You may not see documentation for all members, since class X is not pre-loaded."

But the griffe message already seems like it may be good enough

pawamoy:
This needs a bit more work for proper integration within mkdocstrings' Python handler. I'll explain later.

pawamoy:
Ready for review, if some of you would like to run a final pass :slightly_smiling_face: 
The docs are not perfect, but don't worry, usage through mkdocstrings will be documented in mkdocstrings-python's own docs, with actual examples.

Issue #173, opened 2023-06-26
Additional support for numpy returns section with no annotation, or just type annotation


machow:
**Describe the bug**

For a numpy returns section, it seems like there may be some small issues with two cases:

* when there is no name or type annotation specified (the annotation becomes `":"`)
* when there is only an annotation (the annotation starts with `": "`)

Here's a docstring with comments trying to explain the two cases:

```
        Returns
        -------
        list of int
            A list of integers.
        flag : bool
            Some kind
            of flag.
        x :
            Name only

        # new case ----
        # worked in a previous version of griffe. For functions with a single returns, this is
        # useful for having the return value be the functions return type hint
        :
            No name or annotation

        # new case ----
        # it looked like this may have included ": " in the annotation
        : int
            Only annotation
```

I can start a PR in a second, and definitely am happy to make any tweaks / changes! I realize the spec for the Returns section is a bit ambiguous, but the "no name or annotation" case seems particularly useful for just documenting the return type hint (maybe an issue similar in spirit to this one!: https://github.com/mkdocstrings/griffe/issues/137)

pawamoy:
Yeah, the Numpydoc spec comes short on these cases. I'm OK with supporting them. We should reach out to the style spec maintainers at some point.

PR #174, opened 2023-06-26
fix: numpy returns handles more name, type cases

machow:
Addresses #173 . I'm not particularly good and regexes, so definitely feel free to take this as more of a demonstration of the issue. Happy to make any tweaks / changes :)


src/griffe/docstrings/numpy.py
```diff
@@ -217,7 +217,9 @@ def _read_block(docstring: Docstring, *, offset: int) -> tuple[str, int]:
         |  # or
         (?P<name>{_RE_NAME})\s*:\s*  # just name
         |  # or
-        (?P<type>{_RE_TYPE})\s*  # just type
+        \s*:$  # no name, no type
+        |
+        :?\s*(?P<type>{_RE_TYPE})\s*  # just type
```
pawamoy, line 222: Not tested, but this would look better to me:

```suggestion
        (?::\s*)?(?P<type>{_RE_TYPE})\s*  # just type
        |  # or
        :\s*  # no name, no type
```

The difference is: we don't allow leading spaces before the colon, but allow trailing spaces after it.

tests/test_docstrings/test_numpy.py
```diff
@@ -367,6 +373,21 @@ def test_returns_section(parse_numpy: ParserType) -> None:
         DocstringReturn(name="", annotation="bool", description="Some kind\nof flag."),
     )
 
+    assert_element_equal(
+        sections[0].value[2],
+        DocstringReturn(name="", annotation=None, description="Name only"),
```
pawamoy, line 378: Shouldn't it pick the name here?

```suggestion
        DocstringReturn(name="x", annotation=None, description="Name only"),
```

pawamoy (commented):
Thanks @machow for all your contributions, you're awesome :smile: 
A few comments :slightly_smiling_face:

src/griffe/docstrings/numpy.py
machow: Did a bit of tinkering just now, and I think because the `_RE_TYPE` regex is `.+`, the "just type" regex ends up capturing the "no name, no type" cases.

I think we may need the "no name, no type" regex first, since it's a subset of things our "just type" captures.

tests/test_docstrings/test_numpy.py
machow: Ahhhh, assert_element_equal doesn't test name 😬. Going through now to see how to work it in

src/griffe/docstrings/numpy.py
pawamoy: Ha, sorry for making you do regex gymnastics! I thought my suggestion covered it, but if it's easier to have the  "nothing" case before the "just type" one, then please keep them in that order! I prefer the regex to stay simple and not individually cover badly-supported cases from composed ones. Each regex deals with its own issues :slightly_smiling_face:

machow:
@pawamoy, alright, I think assert_element_equal should be checking all the attributes now. Happy to make any other useful changes

pawamoy:
I'm waiting for an answer at https://github.com/numpy/numpydoc/issues/463. Numpydoc maintainers don't particularly like the `:` and `: int` syntax. If we can't find a better alternative, we'll go ahead with it :slightly_smiling_face:

pawamoy:
Lets wait a few more days, and then proceed to merge. I don't think the discussion on numpydoc will yield immediate results anyway. I'm OK with driving changes rather than waiting for them :slightly_smiling_face:

pawamoy (approved):
OK, LGTM now, what do you think? I'm ready to squash and merge :slightly_smiling_face:

pawamoy:
Merging, thanks again @machow !

Issue #172, opened 2023-06-26
Editable installs broken for pth files using ";" instead of newlines


machow:
**Describe the bug**

Currently, griffe uses a regex over pth files that assumes each command is separated by a line break. I'm pretty unfamiliar with pth files, but ran into some issues with editable installs recently, which seem to trace back to pth files using ";" to terminate lines :/.

Here is a pth file I noticed causing issues:

```
import __editable___quartodoc_0_3_5_dev1_gdba857e_d20230626_finder; __editable___quartodoc_0_3_5_dev1_gdba857e_d20230626_finder.install()
```

I noticed that...

* running through `griffe.finders._handle_pth_file` returned an empty list
* replacing the `;` with a newline caused it to return the two entries pointing to the editable source directory.

I don't know a lot about pth files, but wonder if this is a recent setuptools tweak or something? 😓

pawamoy:
From the docs at https://docs.python.org/3/library/site.html: "Lines starting with import (followed by space or tab) are executed". Since the docs didn't mention semi-colons, we changed the previous split on `;` to a split on `\n` in this commit: https://github.com/mkdocstrings/griffe/commit/f212dd3b92f51a64795fdbb30aefd0a730393523. But I guess it's just ambiguous and semi-colons should be supported in executable lines.

So, a fix could be:

```python
for line in path.read_text(encoding="utf8").strip().replace(";", "\n").splitlines(keepends=False):
```

...hoping that nobody puts `;` in the name of directories added to sys.path :smiling_face_with_tear: 

Would you like to submit a PR? This fix plus a test would be perfect.

PR #175, opened 2023-06-27
fix: parse pth files with semi-colons as breaks

machow:
Addresses #172, by supporting semi-colons in pth files. Thanks for your quick response in the issue! Happy to make any changes :)


pawamoy (approved):
Great, thanks!

PR #187, opened 2023-07-25
feat: Add `--version` and `-V` cli flag.

### LINKED ISSUE #186: Add version CLI flag.

It is not easy to know what version of griffe I am using from the command line interface.

I'd like to be able to do `griffe --version` and get the version information.

jgarte:
Hi,

This PR adds `--version` and `-V` command line flags to `griffe`.

Should I also add the version information to the common options?

This PR also closes #186.

I also added two commits fixing a typo in a docstring and improving the return description.

## Screenshot

image

### DIFF
--- a/src/griffe/cli.py
+++ b/src/griffe/cli.py
@@ -19,6 +19,7 @@ import logging
 import os
 import sys
 from datetime import datetime, timezone
+from importlib.metadata import version as get_version
 from pathlib import Path
 from typing import IO, TYPE_CHECKING, Any, Callable, Sequence
 
@@ -33,6 +34,8 @@ from griffe.git import _get_latest_tag, _get_repo_root, load_git
 from griffe.loader import GriffeLoader, load
 from griffe.logger import get_logger
 from griffe.stats import _format_stats
+
+version = get_version("griffe")
 
 if TYPE_CHECKING:
     from griffe.extensions import Extensions, ExtensionType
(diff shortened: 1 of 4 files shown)


src/griffe/__init__.py
```diff
@@ -16,6 +16,8 @@
 from griffe.loader import load
 from griffe.logger import get_logger
 
+__version__ = "0.32.3"
```
pawamoy, line 19: I'm not a fan of hardcoding the version in the package itself.
Could we use `importlib.metadata.version` instead?
See example in my next comment.

src/griffe/cli.py
```diff
@@ -24,6 +24,7 @@
 
 import colorama
 
+from griffe import __version__
```
pawamoy, line 27: If we use importlib instead:

```python
from importlib.metadata import version as get_version

version = get_version("griffe")
```

pawamoy (changes requested):
Thanks for the PR!

See my comments. If you think it's better to hardcode the version (I know it has benefits), then we should also automate its update through PDM: https://pdm-backend.fming.dev/metadata/#writing-dynamic-version-to-file.

src/griffe/cli.py
jgarte: Hi,

Thanks for the review. Sounds good to me.I added this suggestion.

src/griffe/__init__.py
jgarte: `importlib.metadata.version` sounds good to me. Added. Thanks for the example. Much appreciated!

jgarte:
Hi, is there anything I should change/update in this PR that you haven't mentioned yet?

pawamoy:
Didn't get time to review again, will do tomorrow :slightly_smiling_face: 
Thanks for the updates!

pawamoy (approved):
Sometimes packages are installed without metadata information, so I guard against that by delaying retrieval of the version, and fallback on a 0.0.0 version. Let me know what you think :slightly_smiling_face:

Issue #185, opened 2023-07-24
GriffeLoader breaks dealing with hidden .venv folder


rsenseman:
**Describe the bug**

During my CI a `.venv` folder is created locally to run tests. The presence of this `.venv` folder then breaks the docs generation.

When I attempt to run `mkdocs build` when the `.venv` folder is present, I get the following error message:

```
ERROR    -  Error reading page '<redacted filepath>/<redacted filename>.md': ''
Traceback (most recent call last):
  File "/Users/rsenseman/Development/<redacted project name>/docs/.venv/lib/python3.8/site-packages/griffe/loader.py", line 550, in _get_or_create_parent_module
    parent_module = parent_module[parent_part]
  File "/Users/rsenseman/Development/<redacted project name>/docs/.venv/lib/python3.8/site-packages/griffe/mixins.py", line 31, in __getitem__
    return self.members[parts[0]][parts[1:]]  # type: ignore[attr-defined]
KeyError: ''
```

This is unintuitive because the error message makes it look like there is some issue with the legitimate file `<redacted filename>.md`, but in reality what is breaking is some code in the `GriffeLoader` class that as far as I can tell is attempting to built out the directory structure of the repo.

Here's what the file structure of the Python project looks like with this `.venv` folder:
<img width="321" alt="Screenshot 2023-07-24 at 4 35 41 PM" src="https://github.com/mkdocstrings/pytkdocs/assets/13340505/0b206906-90c3-4af6-847f-0a85b858e29c">

I'm not exactly sure why the `.env` folder specifically causes an issue, but regardless hidden/dot folder should just be ignored. I tried to use the [mkdocs-exclude](https://github.com/apenwarr/mkdocs-exclude) plugin to see if the exclusions would work with `mkdocstrings` but that was unsuccessful.

**Debug logs**
Relevant logs with sensitive project information excluded:
```
ERROR    -  Error reading page '<redacted filepath>/<redacted filename>.md': ''
Traceback (most recent call last):
  File "/Users/rsenseman/Development/<redacted project name>/docs/.venv/lib/python3.8/site-packages/griffe/loader.py", line 550, in _get_or_create_parent_module
    parent_module = parent_module[parent_part]
  File "/Users/rsenseman/Development/<redacted repo name>/docs/.venv/lib/python3.8/site-packages/griffe/mixins.py", line 31, in __getitem__
    return self.members[parts[0]][parts[1:]]  # type: ignore[attr-defined]
KeyError: ''
```

[...]

PR #188, opened 2023-07-26
fix: ignore hidden folders

rsenseman:
### Summary

When searching module subpaths to load files for parsing, skip hidden folders that start with the `.` character. This will prevent unexpected files from being included in the docs generation process

Fixes Issue #185: https://github.com/mkdocstrings/griffe/issues/185


pawamoy:
Thanks @rsenseman for the PR, unfortunately this fixes an unsupported use-case, and wouldn't ignore other unrelated directories whose names don't start with a dot. Details in the issue you mention :slightly_smiling_face:

PR #191, opened 2023-07-30
Add `__post_init__` to places to look for attribute docstrings

### LINKED ISSUE #190: Add __post_init__ as a place to look for attribute docstrings (labeling them as instance-attributes)
I've converted many of my classes to dataclasses. I'd like to add docstrings to many of the attributes that are now added in `__post_init__` rather than `__init__` (since that's really where args/kwargs now live)

Add `__post_init__` to two places in visitor.py that look for attribute docstrings

**Describe alternatives you've considered**
I looked at creating an extension, but was struggling to see how to make it work

ssweber:
With python dataclasses, `__post_init__` essentially becomes an extension of `__init__`, and I'd like to be able to add docstrings to attributes there too.

Closes #190

### DIFF
--- a/src/griffe/agents/visitor.py
+++ b/src/griffe/agents/visitor.py
@@ -475,7 +475,7 @@ class Visitor:
 
         self.extensions.call("on_instance", node=node, obj=function)
         self.extensions.call("on_function_instance", node=node, func=function)
-        if self.current.kind is Kind.CLASS and function.name == "__init__":
+        if self.current.kind is Kind.CLASS and function.name in ["__init__","__post_init__"]:
             self.current = function  # type: ignore[assignment]  # temporary assign a function
             self.generic_visit(node)
             self.current = self.current.parent  # type: ignore[assignment]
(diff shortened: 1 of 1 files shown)


ssweber (commented):
putting:

```
                    # forward value from init, rather than post_init, to preserve behavior
                    value = existing_member.value
```
in 628 would fix the 'overwrite init` if you set it again in post_init, but I'm not sure if that's the real fix.

pawamoy:
I'd prefer to keep the minimum amount of logic regarding `__post_init__`, since IMO in the end we won't look for attribute docstrings into it, but rather extract useful info from `field`s.

ssweber:
Now that I understand the plan, maybe its best to close this pr. My dinky project will survive a few things labeled as `class-variables` that arnt 😄

I can migrate my dataclasses to use field(init=False) for things I set in `__post_init__`. That way, once more logic is extracted from fields, and `__post_init__` handling is deleted, everything will be fixed without further modification.

pawamoy:
OK, sounds good to me :slightly_smiling_face: Closing then! By the way I'm very close to finish my "expressions" improvement, so it shouldn't take long before we're able to extract more data from fields :slightly_smiling_face:

ssweber:
Thanks again for explaining things for me!

PR #209, opened 2023-09-12
Warnings are different from Warns

AntoineD:
In numpy format,  contrary to `Warns`  sections, `Warnings` sections [do not have multiple items](https://numpydoc.readthedocs.io/en/latest/format.html#warnings).

The google format does not specify neither the `Warnings` nor the `Warns` section, but sphinx can parse it like for the numpy format.

This MR distinguish those sections such that `Warnings` sections have no items and docstrings that can be parsed by sphinx can also be parsed by mkdocs.

### DIFF
--- a/docs/docstrings.md
+++ b/docs/docstrings.md
@@ -488,7 +488,6 @@ depending on what you imported/defined in the current module.
 
 #### Warns
 
-- Aliases: Warnings
 - Multiple items allowed
 
 Warns sections allow to document warnings emitted by the following code.
@@ -512,6 +511,19 @@ TIP: **Warnings names are resolved using the function's scope.**
 You can document custom warnings, using the names available in the current scope,
 for example `my_warnings.MyCustomWarning` or `MyCustomWarning` directly,
 depending on what you imported/defined in the current module.
+
+#### Warnings
+
+Warnings sections allow to document cautions for the user.
+
+```python
+def foo():
+    """Foo.
+
+    Warnings:
+        Use this function with caution.
+    """
+```
 
 #### Yields
 
@@ -1219,7 +1231,6 @@ depending on what you imported/defined in the current module.
 
 #### Warns
 
-- Aliases: Warnings
 - Multiple items allowed
 
 Warns sections allow to document warnings emitted by the following code.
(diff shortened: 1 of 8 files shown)


pawamoy:
I initially didn't add support for such Warnings sections because they're just markup (you'll a comment explaining this in the `numpy.py` module docstring). This one is basically equivalent to a warning admonition (though there's no concept of admonitions in our Numpydoc parser).

I'm not sure if I want to keep fighting the docstring standards and their discrepancies. On one hand I don't feel like they will ever evolve, so I should just go with it, and on the other hand I really don't enjoy maintaining all this boilerplate (one enumeration value, four map entries, two functions, two classes) for sections that are just markup. Maybe this whole parsing system could be factorized and improved to get rid of the boilerplate :confused: 

Or maybe we could bring support for Numpydoc admonitions, allowing to have Warning or Note sections that are parsed as admonitions, removing the need for the same, plain markup Numpydoc sections (Warnings, Notes, References).

Or maybe we could mix both things: instead of creating new classes, have the `_read_warnings` functions just return admonitions of type `warning`. It's a bit annoying to lose the "Warnings" alias for "Warns" sections, but maybe that's worth it.

Eager to hear your thoughts on this :slightly_smiling_face:

AntoineD:
By admonition, are you referring to [this](https://squidfunk.github.io/mkdocs-material/reference/admonitions/)?

pawamoy:
Yes :slightly_smiling_face:

AntoineD:
I was not aware of those admonitions (I'm relatively new to mkdocs ecosystem btw).
You make fair points, these make sense.

I feel like projects using numpydoc `Warnings` sections should be either supported as for sphinx or informed that they must convert those sections to warning admonition markup. In any case, I would remove the alias for `Warns` which I find confusing.

At least I think I would rather go the easy path to not support `Warnings` and require to use warning admonition markup.

pawamoy:
Thanks for your feedback.

I think I'll add support for admonitions in the Numpydoc parser, removes the "Warnings" alias for actual "Warns" sections, and tweak things a bit so that Notes is properly rendered as `note` admonition, and Warnings is properly rendered as a `warning` admonition, etc.

AntoineD:
Sounds good, let's close this PR.

PR #210, opened 2023-09-17
feat[docstrings/sphinx]: Warn when unknown params are in docstring.

### LINKED ISSUE #64: docstrings/sphinx: Support "warn unknown params" feature
When a docstring is parsed, and a parameter is found, and the docstring has a parent function attached, we check that the parameter actually exists in the signature.

The following example should emit a warning:

```python
def increment(a: int) -> int:
    """Increment a number.

    :param b: The number.
    """
    return a + 1
```

ashwinvin:
- Added the feature and test for it
- Closes #64


pawamoy:
Thanks! Could you add an option to the Sphinx parser to allow users to enable/disable this behavior?

ashwinvin:
Sure!

pawamoy (approved):
LGTM now. Thanks for the PR!

Issue #215, opened 2023-10-06
fuzzing: Failing cases found with pysource-codegen


pawamoy:
https://github.com/15r10nk/pysource-codegen :rocket: 

```
FAILED tests/test_visitor.py::test_visit_arbitrary_code[115011] - AttributeError: 'str' object has no attribute 'canonical_path'
FAILED tests/test_visitor.py::test_visit_arbitrary_code[184211] - SystemError: compiler_lookup_arg(name='name_5') with reftype=3 failed in <generic parameters of name_5>; freevars of code name_5: ('.type_params', 'name_0', 'name_2', 'name_3', 'name_...
FAILED tests/test_visitor.py::test_visit_arbitrary_code[904162] - AttributeError: 'str' object has no attribute 'canonical_path'
FAILED tests/test_visitor.py::test_visit_arbitrary_code[621041] - SystemError: compiler_lookup_arg(name='name_5') with reftype=3 failed in <generic parameters of name_5>; freevars of code name_5: ('.type_params', 'name_1', 'name_2', 'name_5')
FAILED tests/test_visitor.py::test_visit_arbitrary_code[243509] - SystemError: compiler_lookup_arg(name='name_0') with reftype=2 failed in <generic parameters of name_5>; freevars of code name_5: ('.type_params', 'name_0', 'name_3', 'name_4')
FAILED tests/test_visitor.py::test_visit_arbitrary_code[402833] - AttributeError: 'str' object has no attribute 'canonical_path'
FAILED tests/test_visitor.py::test_visit_arbitrary_code[451286] - AttributeError: 'str' object has no attribute 'canonical_path'
FAILED tests/test_visitor.py::test_visit_arbitrary_code[79708] - SystemError: compiler_lookup_arg(name='name_0') with reftype=3 failed in <generic parameters of name_3>; freevars of code name_3: ('.type_params', 'name_0')
FAILED tests/test_visitor.py::test_visit_arbitrary_code[874497] - AttributeError: 'str' object has no attribute 'canonical_path'
FAILED tests/test_visitor.py::test_visit_arbitrary_code[920242] - SystemError: compiler_lookup_arg(name='name_2') with reftype=3 failed in <generic parameters of name_0>; freevars of code name_0: ('.type_params', 'name_1', 'name_2', 'name_4', 'name_5')
[...]

PR #216, opened 2023-10-06
test: try to parse randomly generated code

15r10nk:
Use pysource-codegen and pysource-minimize combined to find source code which can not be parsed.

some results:

``` python
@'some const text'[name_3]
async def name_5(): # type: some text
    pass
```

``` python
@lambda: name_5
class name_5:
    name_4 = name_0
```

``` python
@1[name_0]
async def name_3(): # type: ignoresome text
    pass
```

It might be that these examples are not as minimal as they could be (the type strings are propably not important). But I think that they are a good starting point.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,6 +91,8 @@ tests = [
     "pytest-xdist>=2.4",
     "hypothesmith>=0.2.0",
     "jsonschema>=4.17.3",
+    "pysource-codegen>=0.4.1",
+    "pysource-minimize>=0.4.0",
 ]
 typing = [
     "mypy>=0.910",
--- a/tests/test_random_code.py
+++ b/tests/test_random_code.py
@@ -1,0 +1,33 @@
+
+
+
+from griffe.tests import temporary_visited_module
+from pysource_codegen import generate
+from pysource_minimize import minimize
+import pytest
+
+@pytest.mark.parametrize("seed",range(200))
+def test_visit_arbitrary_code(seed):
+    code=generate(seed)
+    
+    def contains_bug(code):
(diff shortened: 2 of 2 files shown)


pawamoy:
Oh, thanks a lot! I already have written a `fuzz` task though, let me push it so you can see :sweat_smile: 
It doesn't minimize code though, so we might be able to reuse that part :slightly_smiling_face:

pawamoy:
Have a look at https://github.com/mkdocstrings/griffe/blob/main/duties.py#L302 :slightly_smiling_face:

15r10nk:
No problem. You don't have to merge this merge-request. I wanted to show what I mean with the minimization part. I hope it will help you to fix some bugs.

pawamoy:
Yes I already fixed 3 bugs thanks to pysource-codegen :heart: :rocket: 
And thanks for the PR, I didn't think we could use pysource-minimize like this, I will integrate it in my fuzz task!

15r10nk:
Great :+1:

The SystemErrors are from cpython itself https://github.com/python/cpython/issues/109219 It really looks for me like you are using a pre-release. 
Could you please verify this. 

We should report a bug if this is not the case.

pawamoy:
Yep, I [answered on the issue](https://github.com/mkdocstrings/griffe/issues/215#issuecomment-1751411606), I was indeed using 3.12.0rc1. I've upgraded to 3.12.0, ran many and there's no SystemError anymore.

pawamoy:
Sorry for the notifications, I was just testing something. I reused your code in another commit, marking you as co-author. Thanks again!

PR #220, opened 2023-10-18
fix: Consider space-only lines to be empty

pawamoy:
Related to PR #219: https://github.com/mkdocstrings/griffe/pull/219

@machow

### DIFF
--- a/docs/docstrings.md
+++ b/docs/docstrings.md
@@ -832,10 +832,6 @@ The parser accepts a few options:
     These flags are used to alter the behavior of [doctest][] when testing docstrings,
     and should not be visible in your docs. Default: true.
 - `warn_unknown_params`: Warn about parameters documented in docstrings that do not appear in the signature. Default: true.
-- `allow_section_blank_line`: Allow blank lines in sections' content.
-    When false, a blank line finishes the current section.
-    When true, single blank lines are kept as part of the section.
-    You can terminate sections with double blank lines. Default: false.
 
 #### Attributes
 
--- a/src/griffe/docstrings/google.py
+++ b/src/griffe/docstrings/google.py
@@ -110,7 +110,11 @@ def _read_block_items(docstring: Docstring, *, offset: int, **options: Any) -> I
     while new_offset < len(lines):
         line = lines[new_offset]
 
-        if line.startswith(indent * 2 * " "):
+        if _is_empty_line(line):
+            # empty line: preserve it in the current item
+            current_item[1].append("")
+
+        elif line.startswith(indent * 2 * " "):
             # continuation line
             current_item[1].append(line[indent * 2 :])
 
@@ -124,10 +128,6 @@ def _read_block_items(docstring: Docstring, *, offset: int, **options: Any) -> I
                 f"Confusing indentation for continuation line {new_offset+1} in docstring, "
                 f"should be {indent} * 2 = {indent*2} spaces, not {cont_indent}",
             )
-
-        elif _is_empty_line(line):
-            # empty line: preserve it in the current item
-            current_item[1].append("")
 
         elif line.startswith(indent * " "):
             # indent equal to initial one: new item
--- a/src/griffe/docstrings/numpy.py
+++ b/src/griffe/docstrings/numpy.py
@@ -119,7 +119,16 @@ def _read_block_items(
     while new_offset < len(lines):
         line = lines[new_offset]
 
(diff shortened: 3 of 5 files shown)


pawamoy:
Seems like there's a single issue:

```python
    docstring = """
        Examples
        --------
        Line 1.

        Line 2.
    """
    sections, _ = parse_numpy(docstring, allow_section_blank_line=False)
    assert len(sections) == 2
```

This fails.

machow:
> if line is indented, consider it's a continuation line (this condition catches lines composed only of spaces!)

It seems like this won't be routed to right place, since `_is_empty_line()` uses `.strip()`

pawamoy:
I'm not sure what you mean in the last comment. And I'm not sure my own comment was clear :sweat_smile: 

I'll try to restate things more clearly.

IMO it looks like the `allow_section_blank_line` option does not work correctly.
With this docstring:

```
Examples
--------
Line 1.

Line 2.
```

...and `allow_section_blank_line=False`, it should parse as two sections, one example section and one text section. But it parses as a single example section.

Then, in parameter description (or any other kind of item, like raised exceptions or emitted warnings), the option is currently "baked-in": whatever its value, a single blank line does not break the item or section, and two blank lines break the item and section.

I'd say this behavior makes sense, but this can be debated. I don't see why anyone would use two blank lines instead of one if they don't want to break a section... unless they use Markdown extensions that insert dynamic contents in the docstring, with unchecked leading/trailing new lines?

Maybe the `allow_section_blank_line` should also have an effect on item descriptions as well as sections?

Damn, that sounds so complicated for a single docstring parser :sweat_smile:

machow:
As far as I know, the notion of ending a section based on linebreaks only applies to certain contexts in the parser:

* `_read_block()`: **empty lines can't trigger a new section**. A section ends when a new one is detected. This is because numpydoc sections like Examples can contain a lot of free text, including many empty lines.

* `_read_block_items()`: empty lines can trigger a new section (due to the original parser implementation; it makes sense you need to decide when the parameter items stop). In your original implementation, though, a line starting with 4 spaces is not considered empty (in the blame, [the 4 space rule is from 2 years ago](https://github.com/mkdocstrings/griffe/blame/bc4fa4e1d59133baf654fdbfb5464c72b028c6e5/src/griffe/docstrings/numpy.py#L122)). This is because 4 spaces is a way to indicate that a parameter item description is still going.

## Examples

Since `_read_block()` doesn't use empty lines to trigger new sections, the example docstring you shared should be a single section:

```
Examples
--------
Line 1.

Line 2.
```

Since `_read_block_items()` originally allowed starting a line with 4 spaces to indicate a line is not empty, this was originally read as a single section:

```
Parameters
------------
x: int
    a description
    # <- note the 4 spaces, but pretend this comment doesn't exist
    # <- note the 4 spaces, but pretend this comment doesn't exist
    more description
```

The change `allow_section_blank_line` made was to allow this to be a single section:

```
Parameters
-----------
x: int
    the x description
# <- note no spaces, but pretend this comment doesn't exist
y: int
    the y description
```


## I could be very wrong here!

I've tried to read through the git blame for the parser to get a feel for the original behavior, but I could be missing something 😅. Overall, though, I think the PR may be implementing new, more aggressive terminating of sections, and with numpydoc empty lines are often not meant to signal that a section is ending.

Thanks for taking the time read through all this, and improve the numpy parser! I'm happy to dig deeper where useful!

pawamoy:
Thanks for the feedback. It all makes sense. It guess it boils down to: do we want to allow prose in between sections? Which the Numpydoc style guide does not explicitly allow. If not, then indeed we should *not* check blank lines in `_read_block`, only in `_read_block_items`.

However I'm strong on "line with spaces only is an empty/blank line", because linters will report these lines, and formatters will remove the spaces, and therefore users shouldn't have to rely on those.

machow:
Hey, sorry for the wait -- I tried out a bunch of docstrings in sphinx, and also afaict doesn't care if you have a line that's just 4 spaces. Let me test a little bit more, but I think I'm coming around to...

> However I'm strong on "line with spaces only is an empty/blank line"

pawamoy:
How are your tests going :slightly_smiling_face:?

pawamoy:
Hey @machow, just letting you know that I'll move forward with [this](https://github.com/mkdocstrings/griffe/pull/220#issuecomment-1779678310).

PR #223, opened 2023-10-28
fix super().__init__() calls in docs/extensions.md

johnnoone:
`super().__init__()` does not take a self parameter


pawamoy (approved):
Good catch, thanks a lot!

Issue #218, opened 2023-10-10
API checks: Improve reporting for attribute who lost their value


pawamoy:
**Describe the bug**
For example, going from:

```python
class A:
    b: int | None = None
```

to

```python
class A:
    b: int
```

will report

```
src/package/module.py:2: A.b: Attribute value was changed: None -> None
```

I think it should say "unset" instead of None?

PR #225, opened 2023-11-10
fix: reporting the attribute who lost their value

gpuligundla:
This PR Address the #218 issue. 

Describe the bug
For example, going from:

```
class A:
    b: int | None = None
```
to

```
class A:
    b: int
```
will report

`Attribute value was changed: None -> None`

Now modified to 

`Attribute value was changed: None -> unset`

### DIFF
--- a/src/griffe/diff.py
+++ b/src/griffe/diff.py
@@ -417,7 +417,8 @@ def _attribute_incompatibilities(old_attribute: Attribute, new_attribute: Attrib
     if old_attribute.value != new_attribute.value:
         if new_attribute.value is None:
             yield AttributeChangedValueBreakage(new_attribute, old_attribute.value, "unset")
-        yield AttributeChangedValueBreakage(new_attribute, old_attribute.value, new_attribute.value)
+        else:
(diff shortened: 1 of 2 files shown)


pawamoy (changes requested):
Thanks!

It would be better to pass "unset" to the breakage instance rather than modifying the value in the actual attribute 🙂

gpuligundla:
> Thanks!
> 
> It would be better to pass "unset" to the breakage instance rather than modifying the value in the actual attribute 🙂

Changed it

pawamoy (commented):
Thanks. I realise it needs an `else` clause, otherwise it would yield both breakages.
Could you also add a test :) ?

gpuligundla:
> Thanks. I realise it needs an `else` clause, otherwise it would yield both breakages. Could you also add a test :) ?

ah! yeah sorry for being noob, i assumed it as return keyword

pawamoy (approved):
No worries! And thanks for the update :slightly_smiling_face: LGTM now, merging!

PR #236, opened 2024-02-10
Add special case for NumpyDoc warnings section so they render as warning admonitions

parafoxia:
This PR fixes a bug in the NumpyDoc parser where the Warnings section would render as a note admonition. They now render as warning admonitions.

I felt bad opening an issue about this after accusing the NumpyDoc parser of being buggy yesterday, so thought I'd at least contribute this one myself 😄

### DIFF
--- a/tests/test_docstrings/test_numpy.py
+++ b/tests/test_docstrings/test_numpy.py
@@ -186,7 +186,7 @@ def test_isolated_dash_lines_do_not_create_sections(parse_numpy: ParserType) -> 
     assert sections[1].value.description == "Note contents.\n\n---\nText."
 
 
-def test_admonition_warnings_section(parse_numpy: ParserType) -> None:
+def test_admonition_warnings_special_case(parse_numpy: ParserType) -> None:
     """Test that the "Warnings" section renders as a warning admonition.
 
     Parameters:
@@ -208,6 +208,30 @@ def test_admonition_warnings_section(parse_numpy: ParserType) -> None:
     assert sections[1].title == "Warnings"
     assert sections[1].value.description == "Be careful!!!\n\nmore text"
     assert sections[1].value.annotation == "warning"
+
+
+def test_admonition_notes_special_case(parse_numpy: ParserType) -> None:
+    """Test that the "Warnings" section renders as a warning admonition.
+
+    Parameters:
+        parse_numpy: Fixture parser.
+    """
+    docstring = """
+    Summary text.
+
+    Notes
+    -----
+    Something noteworthy.
+
+    more text
+    """
+
+    sections, _ = parse_numpy(docstring)
+    assert len(sections) == 2
+    assert sections[0].value == "Summary text."
+    assert sections[1].title == "Notes"
+    assert sections[1].value.description == "Something noteworthy.\n\nmore text"
+    assert sections[1].value.annotation == "note"
 
 
 # =============================================================================================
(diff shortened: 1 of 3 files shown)


parafoxia:
I'm not sure what the deal with the (macos-latest, 3.8) tests is, but I don't think that's my fault? 😅

pawamoy:
Indeed, not sure what these `/var` / `/private/var` shenanigans are.

Regarding the change: I don't like modifying kinds in the generic admonition class. Ideally, Griffe is never aware of how admonition kinds are used, which ones are available, etc. It all depends on the downstream docs system. For all we know, maybe someone has defined a custom `warnings` (plural) admonition that differs from a `warning` (singular) one. Or maybe other docstring styles allow such plural kinds.

With this in mind, I'd prefer that we handle this special case in the Numpy parser itself, rather than in the generic admonition class, and only because Warnings section are actually specified in the style-guide (https://numpydoc.readthedocs.io/en/latest/format.html#warnings).

Looks like `_append_section` would be the right place for that:

```python
def _append_section(sections: list, current: list[str], admonition_title: str) -> None:
    if admonition_title:
        kind = admonition_title.lower().replace(" ", "-")
        if kind == "warnings":
            kind = "warning"
        sections.append(
            DocstringSectionAdmonition(
                kind=admonition_title.lower().replace(" ", "-"),
                text="\n".join(current).rstrip("\n"),
                title=admonition_title,
            ),
        )
    elif current and any(current):
        sections.append(DocstringSectionText("\n".join(current).rstrip("\n")))
``` 

WDYT?

parafoxia:
Wrote this comment, made the changes, then realised I forgot to press send lmao. I like your suggestion though, didn't realise there were NumpyDoc specific sections so deffo better to handle the special cases in those. I've also added a special case for the Notes section as well.

tests/test_docstrings/test_numpy.py
```diff
@@ -186,6 +186,54 @@ def test_isolated_dash_lines_do_not_create_sections(parse_numpy: ParserType) ->
...
+
+    more text
+    """
+
+    sections, _ = parse_numpy(docstring)
+    assert len(sections) == 2
+    assert sections[0].value == "Summary text."
+    assert sections[1].title == "Warnings"
+    assert sections[1].value.description == "Be careful!!!\n\nmore text"
+    assert sections[1].value.annotation == "warning"
```
pawamoy, line 210:
```suggestion
    assert sections[1].value.kind == "warning"
```

tests/test_docstrings/test_numpy.py
```diff
@@ -186,6 +186,54 @@ def test_isolated_dash_lines_do_not_create_sections(parse_numpy: ParserType) ->
...
+
+    more text
+    """
+
+    sections, _ = parse_numpy(docstring)
+    assert len(sections) == 2
+    assert sections[0].value == "Summary text."
+    assert sections[1].title == "Notes"
+    assert sections[1].value.description == "Something noteworthy.\n\nmore text"
+    assert sections[1].value.annotation == "note"
```
pawamoy, line 234:
```suggestion
    assert sections[1].value.kind == "note"
```

Same thing, just more explicit :slightly_smiling_face:

pawamoy (approved):
OK, LGTM! Just two suggestions that I'll commit right away, and we can merge :slightly_smiling_face:

parafoxia:
Leaving the branch up as I need to build the docs from it until the next Griffe version is released.

pawamoy:
I wanted to pack a few more things in the next release, but it's been delayed, so I'll release one today :slightly_smiling_face:

pawamoy:
v0.41.0 released.

Issue #239, opened 2024-02-14
bug: A class that derives from a dataclass should be labelled a dataclass


has2k1:
Here is a failing test.

```python
code = '''
from dataclasses import dataclass

@dataclass
class Base:
    pass

class Derived(Base):
    pass
'''

with temporary_visited_module(code) as module:
    obj = module["Derived"]
    assert "dataclass" in obj.labels
```

For reference, the parallel test in core python is.

```python
from dataclasses import dataclass, is_dataclass

@dataclass
class Base:
    pass

class Derived(Base):
    pass

assert is_dataclass(Derived)
```

pawamoy:
Hmmm, this one is problematic. Consider the following example:

```python
from dataclasses import dataclass

@dataclass
class A:
    a: int

class B(A):
    b: int

@dataclass
class C(B):
    c: int

print(inspect.signature(A))
print(inspect.signature(B))
print(inspect.signature(C))
```

Running it will print:

```
(a: int) -> None
(a: int) -> None
(a: int, c: int) -> None
```

It means that, to build the `__init__` method ourselves during static analysis, we have to list attributes from parent classes that are dataclasses, excluding ones that are not. We tell them apart thanks to the `dataclass` label. If we mark `B` as a dataclass, then we will wrongly add `b` as a parameter to `C`'s `__init__` method. 

If we want to support this, I suppose we will have to use something else than the `dataclass` label to tell classes apart. Maybe something internal, stored in our `extra` metadata... Not sure yet. Any other idea?

has2k1:
Yes checking the decorators. I did not know about it. Thanks for that one, I will change my "is_dataclass" checks to use that.

Issue #238, opened 2024-02-13
bug: Regular classes that inherit dataclass do not get dataclass parameters


has2k1:
Here is failing test case

```python
from griffe.tests import temporary_visited_module

code = """
from dataclasses import dataclass

@dataclass
class Base:
    a: int
    b: str

@dataclass
class Derived1(Base):
    c: float

class Derived2(Base):
    d: float 
"""

with temporary_visited_module(code) as module:
    params1 = list(module["Derived1"].parameters)
    params2 = list(module["Derived2"].parameters)
    assert [p.name for p in params1] == ["a", "b", "c"]
    assert [p.name for p in params2] == ["a", "b"] # failing
```

pawamoy:
Thanks, nice catch. The new approach will fix that too.

Issue #234, opened 2024-02-08
bug: parameter kinds for dataclasses should match the non-dataclass equivalents


has2k1:
Currently, parameters derived from dataclass fields have no `kind` information.

This is mainly for python 3.10+ which have keyword_only specification for dataclass fields. Here are failing test cases.

```python
from griffe.tests import temporary_visited_module

code = """
from dataclasses import dataclass, field, KW_ONLY

# Pair of dataclass and equivalent non-dataclass
# The parameter kinds for each pair should be the same
    
@dataclass
class PointA:
    x: float
    y: float = field(kw_only=True)

class PointA_:
    def __init__(self, x: float, *, y: float): ...

@dataclass
class PointB:
    x: float = field(kw_only=True)
    y: float

class PointB_:
    def __init__(self, y: float, *, x: float): ...

@dataclass
class PointC:
    x: float
    _: KW_ONLY
    y: float

class PointC_:
    def __init__(self, x: float, *, y: float): ...

@dataclass
class PointD:
    _: KW_ONLY
    x: float
    y: float

class PointD_:
    def __init__(self, *, x: float, y: float): ...

@dataclass(kw_only=True)
class PointE:
    x: float
    y: float

class PointE_:
    def __init__(self, *, x: float, y: float): ...
"""

def parameter_kinds(obj):
    return [(p.kind and p.kind.value) or None for p in obj.parameters]

with temporary_visited_module(code) as module:
    def both_parameter_kinds(s):
        return [parameter_kinds(obj) for obj in (module[s], module[f"{s}_"])]

    kindsA, kindsA_ = both_parameter_kinds("PointA")
    kindsB, kindsB_ = both_parameter_kinds("PointB")
    kindsC, kindsC_ = both_parameter_kinds("PointC")
    kindsD, kindsD_ = both_parameter_kinds("PointD")
    kindsE, kindsE_ = both_parameter_kinds("PointE")

    assert kindsA == kindsA_
    assert kindsB == kindsB_
    assert kindsC == kindsC_
    assert kindsD == kindsD_
    assert kindsE == kindsE_
```

PR #240, opened 2024-02-29
feat: Better support for dataclasses

### LINKED ISSUE #233: bug: A dataclass field with init=False should not be a parameter
Here are failing tests.

```python
from griffe.tests import temporary_visited_module

code = """
from dataclasses import dataclass, field

@dataclass
class PointA:
    x: float
    y: float
    z: float = field(init=False)

@dataclass(init=False)
class PointB:
    x: float
    y: float

@dataclass(init=False)
class PointC:
    x: float
    y: float = field(init=True)  # init=True has no effect
"""

with temporary_visited_module(code) as module:
    paramsA = module["PointA"].parameters
    paramsB = module["PointB"].parameters
    paramsC = module["PointC"].parameters

    assert "z" not in paramsA
    assert "x" not in paramsB and "y" not in paramsB
    assert "x" not in paramsC and "y" not in paramsC
```

Then there are the perhaps more complicated cases to deal with statically where the `field` class is wrapped inside another function e.g.

```python
def no_init(default: T) -> T:
    """
    Set default value of a dataclass field that will not be __init__ed
    """
    return field(init=False, default=default)

@dataclass
class Point:
    x: float
    y: float = no_init(0)  # y is not a parameter
```

pawamoy:
Closes #233, #234, #238, #239.

TODO: 

- [x] remove `d.py`
- [x] add more tests (coming from mentioned issues)
- [ ] detect when dataclasses extension functionality was used, to issue a deprecation warning? or leave it enabled by default for eternity?

### DIFF
--- a/src/griffe/extensions/base.py
+++ b/src/griffe/extensions/base.py
@@ -252,9 +252,6 @@ class Extensions:
         self._extensions: list[Extension] = []
         self.add(*extensions)
 
-        # TODO: Deprecate and remove at some point?
-        self.add(*_load_extension("dataclasses"))
-
     def add(self, *extensions: ExtensionType) -> None:
         """Add extensions to this container.
 
--- a/src/griffe/extensions/base.py
+++ b/src/griffe/extensions/base.py
@@ -473,8 +473,17 @@ def load_extensions(
             extensions.add(*ext)
         else:
             extensions.add(ext)
+
     # TODO: Deprecate and remove at some point?
-    extensions.add(*_load_extension("dataclasses"))  # type: ignore[misc]
+    # Always add our built-in dataclasses extension.
+    from griffe.extensions.dataclasses import DataclassesExtension
+
+    for ext in extensions._extensions:
+        if type(ext) == DataclassesExtension:
+            break
+    else:
+        extensions.add(*_load_extension("dataclasses"))  # type: ignore[misc]
+
     return extensions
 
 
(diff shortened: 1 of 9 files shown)


pawamoy:
@has2k1 would you like to review? When squashing I'll add you as co-author since your issues were super helpful and I reused your code for tests :slightly_smiling_face:

src/griffe/extensions/base.py
```diff
@@ -252,6 +252,9 @@ def __init__(self, *extensions: ExtensionType) -> None:
         self._extensions: list[Extension] = []
         self.add(*extensions)
 
+        # TODO: Deprecate and remove at some point?
+        self.add(*_load_extension("dataclasses"))
+
```
pawamoy, line 257: Oops, must remove this.

has2k1:
Nice, looks good so far. I will have time to test it over the weekend.

src/griffe/extensions/base.py
```diff
@@ -465,12 +468,14 @@ def load_extensions(exts: Sequence[str | dict[str, Any] | ExtensionType | type[E
...
     extensions = Extensions()
-    for extension in exts:
+    for extension in exts or ():
         ext = _load_extension(extension)
         if isinstance(ext, list):
             extensions.add(*ext)
         else:
             extensions.add(ext)
+    # TODO: Deprecate and remove at some point?
+    extensions.add(*_load_extension("dataclasses"))  # type: ignore[misc]
```
pawamoy, line 478: Should we add it first or last? Should we check if it is already added by the provided `exts`?

PR #248, opened 2024-03-11
fix: Fix dataclass parameters to exclude properties

has2k1:
This was fixed in 5a5c03b, but broken in 82a9d57. The tests
broke as well and did not catch the regression.


pawamoy (approved):
Awesome, thanks!

PR #252, opened 2024-03-12
fix: Don't turn items annotated as InitVar into dataclass members

has2k1:
This PR addresses the case of dataclass [init-only variables](https://docs.python.org/3/library/dataclasses.html#init-only-variables) also being treated as class members. i.e

```python
from dataclasses import dataclass, InitVar

@dataclass
class Point:
    x: float
    y: float
    z: InitVar[float]
```

is equivalent (compatible) to

```python
class Point:
    def __init__(self, x: float, y: float, z: float):
       self.x = x
       self.y = y
```

The class is not expected to have member `z` (`instance.z`).

### DIFF
--- a/src/griffe/extensions/dataclasses.py
+++ b/src/griffe/extensions/dataclasses.py
@@ -176,6 +176,17 @@ def _set_dataclass_init(class_: Class) -> None:
     class_.set_member("__init__", init)
 
 
+def _del_members_annotated_as_initvar(class_: Class) -> None:
+    # Definitions annotated as InitVar are not class members
+    attributes = cast(
+        list[Attribute],
+        [member for member in class_.members.values() if member.is_attribute]
+    )
+    for attribute in attributes:
+        if isinstance(attribute.annotation, Expr) and attribute.annotation.canonical_path == "dataclasses.InitVar":
+            class_.del_member(attribute.name)
+
+
 def _apply_recursively(mod_cls: Module | Class, processed: set[str]) -> None:
     if mod_cls.canonical_path in processed:
         return
@@ -183,6 +194,7 @@ def _apply_recursively(mod_cls: Module | Class, processed: set[str]) -> None:
     if isinstance(mod_cls, Class):
         if "__init__" not in mod_cls.members:
             _set_dataclass_init(mod_cls)
+            _del_members_annotated_as_initvar(mod_cls)
         for member in mod_cls.members.values():
             if not member.is_alias and member.is_class:
                 _apply_recursively(member, processed)  # type: ignore[arg-type]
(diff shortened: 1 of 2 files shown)


src/griffe/extensions/dataclasses.py
```diff
@@ -176,13 +176,25 @@ def _set_dataclass_init(class_: Class) -> None:
     class_.set_member("__init__", init)
 
 
+def _del_members_annotated_as_initvar(class_: Class) -> None:
+    # Definitions annotated as InitVar are not class members
+    attributes = cast(
+        list[Attribute],
+        [member for member in class_.members.values() if member.is_attribute]
+    )
```
pawamoy, line 184: Unfortunately this doesn't work on Python 3.8 (and 3.9 I think), as types are not subscriptable until 3.10. You can simply remove the cast and use a `# type: ignore[...]` comment.

pawamoy (commented):
I really know nothing about dataclasses haha! Thanks :slightly_smiling_face:

src/griffe/extensions/dataclasses.py
pawamoy: Also, it could be preferable to use `attributes = [member for member in class_.members.values() if isinstance(member, Attribute)`. This way you don't need the type ignore comment or the cast, and we also rule out aliases, since they couldn't possibly be annotated with `InitVar` anyway.

has2k1:
> I really know nothing about dataclasses haha! Thanks 🙂

I have just accepted them as a Whac-A-Mole situation.

tests/test_dataclasses.py
```diff
@@ -276,3 +276,37 @@ class Reordered(Base):
...
+    """
+
+    with temporary_visited_package("package", {"__init__.py": code}) as module:
+        PointA = module["PointA"]
+        assert ["self", "x", "y", "z"] == [p.name for p in PointA.parameters]
+        assert ["x", "y", "__init__"] == [name for name in PointA.members]
+
+        PointB = module["PointB"]
+        assert ["self", "r"] == [p.name for p in PointB.parameters]
+        assert ["x", "y", "z", "__init__"] == [name for name in PointB.members]
```
pawamoy, line 312: Ideally `z` would be removed here too, but since `__init__` was already present we didn't apply the dataclasses logic so `z` wasn't removed. That's fine, lets just leave it like that: we are all consenting adults and users can shoot their own foot if they want to :slightly_smiling_face:

pawamoy (approved):
OK, LGTM, lets squash and merge :slightly_smiling_face:

pawamoy:
Ah wait, there are some linting warnings, let me fix those.

PR #253, opened 2024-03-12
fix: Don't return class variables as parameters of dataclasses

has2k1:


tests/test_dataclasses.py
```diff
@@ -91,16 +91,18 @@ def test_alias_proxies() -> None:
...
     code = """
         from dataclasses import dataclass
         from functools import cached_property
+        from typing import ClassVar
 
         @dataclass
         class Point:
             x: float
             y: float
+            z: ClassVar[float] = 3
```
pawamoy, line 105: What if you add

- `a: float = 0` (default value, triggering the addition of a `class-attribute` label)
- `b: ClassVar[float]` (no default value, but annotated so `class-attribute` label added)

tests/test_dataclasses.py
```diff
@@ -102,7 +102,11 @@ def test_dataclass_properties_and_class_variables() -> None:
         class Point:
             x: float
             y: float
-            z: ClassVar[float] = 3
+
+            # These definitions create class variables
+            r: ClassVar[float]
+            s: float = 3
```
pawamoy, line 108: So the test still passes but `s` should appear in the `__init__` method parameters right?

has2k1: Yeah right!

I suspected an edge case because attribute annotation expression is sliced and the `ClassVar` is removed. It was the more obvious solution. That means the solution should check for `instance-attribute` as-well.

pawamoy (approved):
Thanks a lot!

Issue #286, opened 2024-06-07
feature: A way to get docstrings, ... of InitVar parameters of a dataclass


has2k1:
### History

PR #252 removed `InitVar` parameters from the list of dataclass members (and attributes). This was correct.

### New Problem

There is no way to get access to the docstrings of the `InitVar` parameters. Class parameters (`dc.Parameter`) do not have docstrings, only the core objects (`dc.Object`) do and now classes only keep track of those objects that are members (`dc.Class.members`).

```python
from griffe.tests import temporary_visited_package

code = '''
from dataclasses import InitVar, dataclass

@dataclass
class ClassWithInitVar:
    """
    Class with an InitVar Parameter
    """

    a: InitVar[int] = 1
    "Parameter a"

    b: float = 2
    "Parameter b"

    def __post_init__(self, a: int): ...
'''

with temporary_visited_package("package", {"__init__.py": code}) as m:
    obj = m["ClassWithInitVar"]

print(obj.members) # does not include object a  [correct]
print(obj.all_members) # does not include object a [correct]
# print(obj.*) # Nothing includes object a [missing]
```

### Brainstorming a solution

To be useful, the type of `InitVar` parameters should be derived from  `dc.Object`. Then they have `.docstring`, `.lines`, `.lineno`, e.t.c.

But none of the current subclasses `[Alias, Attribute, Class, Function, Module]` is suitable.

Possible options are:

- `InitVar`
- `InitVariable`
- `InitParameter`
- `ClassParameter`
- `DataclassParameter`
- ...

With this kind of solution, the new class category would have to be used for one of:

1. Only dataclass parameters that are InitVar
2. All dataclass parameters

Option 2. seems like the better choice, but some (most) dataclass parameters are also attributes and that would complicate the solution.

That leaves option 1.

Then where should these "InitVar" objects be stored?

Maybe have them back in `dc.Class.members`. This shouldn't be the problem if they are not of type `dc.Attribute` and therefore do not show up in `dc.Class.attributes`. This would effectively make `dc.Class.members` a lookup for all objects statically defined in the class namespace.

pawamoy:
Hi @has2k1, thanks for the report :slightly_smiling_face: 

I'd be more to open to a fundamental change that adds docstrings to parameters and returned/received/yielded values. This would play well with PEP 727 and our griffe-typingdoc extension supporting it.

[...]

PR #288, opened 2024-06-10
feat: Add docstring and code source information to Parameter

has2k1:
This change gives the Parameter class the following attributes / propeties:

   - docstring
   - lineno
   - endlineno
   - parent
   - lines
   - linescollections
   - source

Currenlty these values are only populated for dataclasses, and they are None for regular classes and functions. This makes it possible to accesss this information for all dataclass parameters, especially those set as `InitVar` where it was not possible to get it since they are not members (attributes) of the class.

Issue #286: https://github.com/mkdocstrings/griffe/issues/286
Related to mkdocstrings/griffe#252: https://github.com/mkdocstrings/griffe/pull/252

### DIFF
--- a/src/griffe/dataclasses.py
+++ b/src/griffe/dataclasses.py
@@ -176,11 +176,8 @@ class Parameter:
         annotation: str | Expr | None = None,
         kind: ParameterKind | None = None,
         default: str | Expr | None = None,
-        lineno: int | None = None,
-        endlineno: int | None = None,
         docstring: Docstring | None = None,
         parent: Module | Class | None = None,
-        lines_collection: LinesCollection | None = None,
     ) -> None:
         """Initialize the parameter.
 
@@ -189,11 +186,8 @@ class Parameter:
             annotation: The parameter annotation, if any.
             kind: The parameter kind.
             default: The parameter default, if any.
-            lineno: The parameter starting line, or None for modules. Lines start at 1.
-            endlineno: The parameter ending line (inclusive), or None for modules.
             docstring: The parameter docstring.
             parent: The object parent.
-            lines_collection: A collection of source code lines.
         """
         self.name: str = name
         """The parameter name."""
@@ -203,15 +197,10 @@ class Parameter:
         """The parameter kind."""
         self.default: str | Expr | None = default
         """The parameter default value."""
-        self.lineno: int | None = lineno
-        """The starting line number of the parameter."""
-        self.endlineno: int | None = endlineno
-        """The ending line number of the parameter."""
         self.docstring: Docstring | None = docstring
         """The parameter docstring."""
         self.parent: Module | Class | None = parent
         """The parent of the parameter (none if top module)."""
-        self._lines_collection: LinesCollection | None = lines_collection
 
     def __str__(self) -> str:
         param = f"{self.name}: {self.annotation} = {self.default}"
@@ -237,59 +226,6 @@ class Parameter:
         """Whether this parameter is required."""
         return self.default is None
 
-    @property
-    def module(self) -> Module:
-        """The parent module of this object.
-
-        Raises:
-            ValueError: When the object is not a module and does not have a parent.
-        """
-        if isinstance(self, Module):
-            return self
-        if self.parent is not None:
-            return self.parent.module
-        raise ValueError(f"Object {self.name} does not have a parent module")
-
-    @property
-    def filepath(self) -> Path | list[Path]:
-        """The file path (or directory list for namespace packages) where this object was defined."""
-        return self.module.filepath
-
-    @property
-    def lines_collection(self) -> LinesCollection:
-        """The lines collection attached to this object or its parents.
-
-        Raises:
-            ValueError: When no modules collection can be found in the object or its parents.
-        """
-        if self._lines_collection is not None:
-            return self._lines_collection
-        if self.parent is None:
-            raise ValueError("no lines collection in this object or its parents")
-        return self.parent.lines_collection
-
-    @property
-    def lines(self) -> list[str]:
-        """The lines containing the source of this object."""
-        try:
-            filepath = self.filepath
-        except BuiltinModuleError:
-            return []
-        if isinstance(filepath, list):
-            return []
-        try:
-            lines = self.lines_collection[filepath]
-        except KeyError:
-            return []
-        if self.lineno is None or self.endlineno is None:
-            return []
-        return lines[self.lineno - 1 : self.endlineno]
-
-    @property
-    def source(self) -> str:
-        """The source code of this object."""
-        return dedent("\n".join(self.lines))
-
     def as_dict(self, *, full: bool = False,  **kwargs: Any) -> dict[str, Any]:  # noqa: ARG002
         """Return this parameter's data as a dictionary.
 
@@ -304,12 +240,11 @@ class Parameter:
             "annotation": self.annotation,
             "kind": self.kind,
             "default": self.default,
-            "lineno": self.lineno,
-            "endlineno": self.endlineno,
         }
         if self.docstring:
             base["docstring"] = self.docstring.as_dict(full=full)
         return base
+
 
 class Parameters:
     """This class is a container for parameters.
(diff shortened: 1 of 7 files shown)


has2k1:
@pawamoy, please kill [this](https://github.com/mkdocstrings/griffe/actions/runs/9449093864) action.

When the [test case](https://github.com/mkdocstrings/griffe/blob/e5572d2eb1dd8dbe8f9b43b33119bd9becc4a4d9/tests/test_encoders.py#L14) for the encoder fails, the json [strings](https://github.com/mkdocstrings/griffe/blob/e5572d2eb1dd8dbe8f9b43b33119bd9becc4a4d9/tests/test_encoders.py#L27) being compared are so large they choke pytest when processing the error.

has2k1:
It is ready review, the latest changes [pass](https://github.com/has2k1/griffe/actions/runs/9450918710) (for python<3.13) on my repo.

pawamoy (changes requested):
Thanks a lot for the PR @has2k1!

So, with these changes, Parameters start to look a lot like Objects :thinking: I'm not sure that's something I want. That creates a bit of code duplication. We could reuse mixins to avoid that, but we don't want to go full in and make Parameters actual Objects, because they're not.

I see that you added these attributes/properties:


-    docstring
-    lineno
-    endlineno
-    parent
-    lines
-    linescollections
-    source

May I know why not just adding a `docstring` attribute (and parameter)?

To me, `source` will not be super useful: most of the time, multiple parameters appear on the same line. Even if we're OK with that, getting the source with these line numbers will most of the time again return code with *invalid syntax* (because the code will be incomplete, for example providing just the signature of a function, or even just a line of it). This is in contrast to the other objects, where their `source` property always return valid/complete code.

If `lineno`, `endlineno`, `lines` and `lines_collection` were added to support `source`, then I think we can just get rid of all of them. I prefer not providing API than providing incorrect one. We can always revise later.

The `parent` attribute is probably interesting to have, just for navigation purposes (going from parameter to the function above). But then, lets keep things simple for now, and lets avoid giving Parameter too much of the shape of Objects: lets remove the `module` and `filepath` properties too. They can be accessed by passing through the parent first.

Finally, I wonder if using actual Docstrings makes sense: I don't think it makes sense to use docstring *sections*, or a particular docstring style for parameter docstrings? They usually are plain markup strings, no? Then, maybe parameter docstrings should just be strings?

has2k1:
You are right, the `docstring` is main information.

> May I know why not just adding a docstring attribute (and parameter)

Using the source lines is a cheap hack to get nicely formatted code snippets, especially to deal with multiline annotations.

As these values are not available for non-dataclass parameters, it does make sense to limit the scope of what is added in this change.

src/griffe/dataclasses.py
```diff
@@ -176,6 +176,8 @@ def __init__(
         annotation: str | Expr | None = None,
         kind: ParameterKind | None = None,
         default: str | Expr | None = None,
+        docstring: Docstring | None = None,
+        parent: Module | Class | None = None,
```
pawamoy, line 180: Our `parent` is most definitely a Function here. Maybe we could rename `parent` to `function` to make this more clear and avoid any ambiguity?

src/griffe/dataclasses.py
```diff
@@ -184,6 +186,8 @@ def __init__(
             annotation: The parameter annotation, if any.
             kind: The parameter kind.
             default: The parameter default, if any.
+            docstring: The parameter docstring.
+            parent: The object parent.
```
pawamoy, line 190: suggestion: The parameter's parent function.

src/griffe/dataclasses.py
```diff
@@ -193,6 +197,10 @@ def __init__(
         """The parameter kind."""
         self.default: str | Expr | None = default
         """The parameter default value."""
+        self.docstring: Docstring | None = docstring
+        """The parameter docstring."""
+        self.parent: Module | Class | None = parent
+        """The parent of the parameter (none if top module)."""
```
pawamoy, line 203: suggestion: The parameter's parent function.

pawamoy (changes requested):
Thanks, it's super clean now :)

Just a few suggestions.

src/griffe/dataclasses.py
has2k1: At the moment, the `parent` is a `Class`, because we are transforming an "Attribute".

pawamoy: But the parameter is attached to the `__init__` function that the dataclasses extension generates for the dataclass, right?

First, Griffe parses the InitVar as a class attribute. Then, the dataclasses extension handles the dataclass by re-creating an `__init__` Function and creating it's parameters based on InitVars and else. The docstring of the attribute is passed to the corresponding parameter, then the InitVar attribute is removed. I think it works that way, or at least it should work that way. I'll check the code again.

In the end, the parent of a parameter can only ever be a Function, not a Class 🤔

has2k1:
```python
from griffe.tests import temporary_visited_package

code = """
from dataclasses import dataclass, InitVar

@dataclass
class Base:
    a: int
    "Parameter a"
    b: InitVar[int] = 3
    "Parameter b"
"""

with temporary_visited_package("package", {"__init__.py": code}) as module:
    base = module["Base"]
    param_a = base.parameters[1]
    param_b = base.parameters[2]

type(param_a.parent), type(param_b.parent)
# (griffe.dataclasses.Class, griffe.dataclasses.Class)
```


The `__init__` function is [created](https://github.com/has2k1/griffe/blob/371318190ecfc5329f020a52e0f156dd91031031/src/griffe/extensions/dataclasses.py#L176-L187) after the parameters. 

At [member.parent](https://github.com/has2k1/griffe/blob/371318190ecfc5329f020a52e0f156dd91031031/src/griffe/extensions/dataclasses.py#L129), [member](https://github.com/has2k1/griffe/blob/371318190ecfc5329f020a52e0f156dd91031031/src/griffe/extensions/dataclasses.py#L80) is an "Attribute".

has2k1: Same thing in the [Visitor.handle_function](https://github.com/has2k1/griffe/blob/371318190ecfc5329f020a52e0f156dd91031031/src/griffe/agents/visitor.py#L362-L387).

has2k1: A solution would be:
1. create the function without the parameters
2. create the parameters and pass the function as the parent
3. set the parameters attribute on the function

```python
function = Function(..., parent=None)
function.parameters = Parameters([Parameter(..., parent=function) ...])
```

pawamoy: Right. Then yes, we could do as you suggest, or do something a bit less pure and assign the function to the parameters when we pass them, just like we do when assigning a member to an object:

```python
# in Function.__init__
for parameter in parameters:
    parameter.function = self
```

src/griffe/dataclasses.py
```diff
@@ -193,6 +195,12 @@ def __init__(
         """The parameter kind."""
         self.default: str | Expr | None = default
         """The parameter default value."""
+        self.docstring: Docstring | None = docstring
+        """The parameter docstring."""
+        # The parent function is set in Function.__init__ when the
+        # parameter(s) are assigned to a function
+        self.parent: Function | None = None
```
pawamoy, line 202: Let's call it `function`.

```suggestion
        # The parent function is set in `Function.__init__`,
        # when the parameters are assigned to the function.
        self.function: Function | None = None
```

src/griffe/dataclasses.py
```diff
@@ -1650,6 +1661,9 @@ def __init__(
         self.overloads: list[Function] | None = None
         """The overloaded signatures of this function."""
 
+        for parameter in self.parameters:
+            parameter.parent = self
```
pawamoy, line 1665:
```suggestion
            parameter.function = self
```

pawamoy (changes requested):
Thanks :slightly_smiling_face: I'll just rename `parent` to `function` and I think we're good.

pawamoy (approved):

pawamoy:
Squashed and merged, thanks again!

PR #297, opened 2024-06-26
fix: Attributes without annotations cannot be parameters

has2k1:
All dataclass parameters have annotations. When determining
which of the attributes of dataclass are parameters, we
explicitly check against that condition. This help us filter
out obvious non-candidates that may otherwise fall through.


pawamoy (approved):
Thanks! LGTM.

Issue #263, opened 2024-05-02
bug: Google docstrings: no support for non-multiple or non-named values in Yields section


the-13th-letter:
### Description of the bug
In Google-style docstrings, Griffe (and mkdocstrings-python) don't treat Yields and Returns sections identically when it comes to returning multiple and/or named items.  In particular, for Returns sections, the `returns_multiple_items` and `returns_named_values` configuration options exist and do as advertised.  For Yields sections, no such dedicated options exist, and *those* options are ignored.  As a consequence, for multiple values, there is no way to document the yielded tuple of values *as a tuple*; it must be faked via a single item tuple, the type repeated (actually: overruled) in the docstring, and the description must be written with hanging indentation.

Note: The Google style guide, at least as of version [8487c08][GOOGLESTYLE_REPO], [treats "Yields:" and "Returns:" identically and requires them to document a tuple of returned values as a tuple, explicitly forbidding the `numpy`-style named tuple return values][GOOGLESTYLE].

[GOOGLESTYLE_REPO]: https://github.com/google/styleguide/commit/8487c083e1faecb1259be8a8873618cfdb69d33d
[GOOGLESTYLE]: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

### To Reproduce

Given the following example module `yielding_test`

```python
"""Test "Yields:" sections in Google-style docstrings."""

from __future__ import annotations

from collections.abc import Iterator

def return_one() -> str:
    """XXX

    Returns:
        Returns one item.  Returns one item.  Returns one item.  Returns one
        item.  Returns one item.  Returns one item.

    """
    return "one"

def return_two() -> tuple[str, int]:
    """XXX

    Returns:
        retval: Returns two items.  Returns two items.  Returns two items.
        Returns two items.  Returns two items.  Returns two items.

    """
    return "one", 2

def yield_one() -> Iterator[str]:
    """XXX

    Yields:
        Yields one item.  Yields one item.  Yields one item.  Yields one
        item.  Yields one item.  Yields one item.

    """
    yield "one"

def yield_two() -> Iterator[tuple[str, int]]:
    """XXX

    Yields:
        yieldval: Yields two items.  Yields two items.  Yields two
        items.  Yields two items.  Yields two items.  Yields two items.

    """
    yield "one", 2
```

[...]

PR #322, opened 2024-09-08
Support non-multiple and non-named values in Yields and Receives

the-13th-letter:
This PR fixes https://github.com/mkdocstrings/griffe/issues/263 by applying the same logic from the `Returns` section parser to the `Yields` and `Receives` parser.  The `Yields` section parser is configured with the same two configuration variables `returns_multiple_items` and `returns_named_value`, whereas the `Receives` section parser is configured via the new `receives_multiple_items` and `receives_named_value` configuration variables. Furthermore, the parsing of multiple or single docstring block contents, of named or unnamed parameters, and the unpacking of generator and/or tuple return annotations has been refactored into separate functions.

---

<b>Contents:</b>

- 8a78e68e7702d5b874daa21aeca58943eb9cf7d6 duplicates the non-multiple return and the named value parsing support for the `Yields` and `Receives` section parsers, essentially via copy-and-paste. Includes suitable tests, based on the `Returns` section tests.
- 756118fc0da4866c80fde0b456cbb497124c0a18 implements the aforementioned refactored functions. This commit is functionally optional, but hopefully useful with respect to future ease of maintenance.
- e01447a5980ca2c6e94c93f8df7618deb59d627a fixes linting errors.
- 8e7cb30fac0df5e619da94b102d5c88de0b36a18 adds documentation for the above features.

### DIFF
--- a/src/_griffe/docstrings/google.py
+++ b/src/_griffe/docstrings/google.py
@@ -521,22 +521,37 @@ def _read_yields_section(
     docstring: Docstring,
     *,
     offset: int,
+    returns_multiple_items: bool = True,
+    returns_named_value: bool = True,
     **options: Any,
 ) -> tuple[DocstringSectionYields | None, int]:
     yields = []
-    block, new_offset = _read_block_items(docstring, offset=offset, **options)
+
+    if returns_multiple_items:
+        block, new_offset = _read_block_items(docstring, offset=offset, **options)
+    else:
+        one_block, new_offset = _read_block(docstring, offset=offset, **options)
+        block = [(new_offset, one_block.splitlines())]
 
     for index, (line_number, yield_lines) in enumerate(block):
-        match = _RE_NAME_ANNOTATION_DESCRIPTION.match(yield_lines[0])
-        if not match:
-            docstring_warning(
-                docstring,
-                line_number,
-                f"Failed to get name, annotation or description from '{yield_lines[0]}'",
-            )
-            continue
-
-        name, annotation, description = match.groups()
+        if returns_named_value:
+            match = _RE_NAME_ANNOTATION_DESCRIPTION.match(yield_lines[0])
+            if not match:
+                docstring_warning(
+                    docstring,
+                    line_number,
+                    f"Failed to get name, annotation or description from '{yield_lines[0]}'",
+                )
+                continue
+            name, annotation, description = match.groups()
+        else:
+            name = None
+            if ":" in yield_lines[0]:
+                annotation, description = yield_lines[0].split(":", 1)
+                annotation = annotation.lstrip("(").rstrip(")")
+            else:
+                annotation = None
+                description = yield_lines[0]
         description = "\n".join([description.lstrip(), *yield_lines[1:]]).rstrip("\n")
 
         if annotation:
@@ -554,7 +569,7 @@ def _read_yields_section(
                     raise ValueError
                 if isinstance(yield_item, ExprName):
                     annotation = yield_item
-                elif yield_item.is_tuple:
+                elif yield_item.is_tuple and returns_multiple_items:
                     annotation = yield_item.slice.elements[index]
                 else:
                     annotation = yield_item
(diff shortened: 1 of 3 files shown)


the-13th-letter:
Concerning code style, I come from a `black`/`ruff` environment, and found no settings in your code repository. So my formatting and linting may be off.

src/_griffe/docstrings/google.py
```diff
@@ -491,22 +548,16 @@ def _read_returns_section(
...
+                orig_annotation = annotation
+                annotation = _maybe_destructure_annotation(
+                    annotation, index, multiple=returns_multiple_items
+                )
+                if annotation == orig_annotation:
+                    annotation = _maybe_destructure_annotation(
+                        _unpack_generators(annotation, 2),
+                        index,
+                        multiple=returns_multiple_items,
+                    )
```
pawamoy: It all looks good, except maybe this part. I find it a bit hard to follow, compared to the previous version.

Since `_unpack_generators` is called without `mandatory=True` here, couldn't we directly do this?

```python
                annotation = _maybe_destructure_annotation(
                    _unpack_generators(annotation, 2),
                    index,
                    multiple=returns_multiple_items,
                )
```

I'll try this locally and see if tests pass.

pawamoy (commented):
Thanks a lot! I appreciate the effort on splitting commits to ease review :slightly_smiling_face:

src/_griffe/docstrings/google.py
```diff
@@ -491,22 +552,18 @@ def _read_returns_section(
...
+                    annotation,
+                    index,
+                    multiple=returns_multiple_items,
+                )
+                if annotation == orig_annotation:
+                    annotation = _maybe_destructure_annotation(
+                        _unpack_generators(annotation, 2),
+                        index,
+                        multiple=returns_multiple_items,
+                    )
```
pawamoy, line 566: Tested locally and tests pass with this change:

```suggestion
                annotation = _maybe_destructure_annotation(
                    _unpack_generators(annotation, 2),
                    index,
                    multiple=returns_multiple_items,
                )
```

I wonder if we're missing something though, but IIRC tests are quite comprehensive?

src/_griffe/docstrings/google.py
```diff
@@ -442,6 +442,76 @@ def _read_warns_section(
...
+
+def _unpack_generators(
+    annotation: Any,
+    generator_pos: int,
+    *,
+    mandatory: bool = False,
+) -> Any:
+    if annotation.is_generator:
+        return annotation.slice.elements[generator_pos]
+    if annotation.is_iterator and generator_pos in {0, 2}:
```
pawamoy, line 495: I'm not entirely convinced about the `generator_pos` check (`generator_pos in {0, 2}`). Raising "must be a generator" because `generator_pos` is > 2 would feel wrong, and we only ever pass 0, 1 and 2 anyway, so maybe we could just drop this check?

src/_griffe/docstrings/google.py
the-13th-letter: It's your call. I translated the code very directly from the non-PR version, and it included this try-unpacking-else-un-generator-then-unpack dance. I *did* get a "organically grown and never pruned" vibe from this part, but I didn't study Griffe's workings and code history deeply enough to be comfortable with pruning this section… at least not without double-checking with you first.

In any case, I don't see any problems with your suggestion, based solely on reading and rewriting that code section.

src/_griffe/docstrings/google.py
the-13th-letter: Looking back at this, I think I should have set up an enum instead, instead of magic numbers. Makes for a cleaner interface, and better type checking support.

src/_griffe/docstrings/google.py
pawamoy: Thanks, that makes sense.

> an "organically grown and never pruned" vibe

Definitely. The parsers were never something I felt interested in (for many reasons), and I just coded them lazily, in an imperative mood. I didn't spend much time trying to make them elegant or efficient, as long as they are working as intended. Our tests suite says the suggestion works, so lets apply it. If it turns out the tests are insufficient, we'll fix that then :relaxed:

src/_griffe/docstrings/google.py
the-13th-letter: Not converting the section to enums, [as requested](https://github.com/mkdocstrings/griffe/pull/322#pullrequestreview-2289652118).

the-13th-letter:
GitHub doesn't seem to recognize a PR that has been reset to an earlier, already-reviewed state as already reviewed…

Also, there's an apparently unrelated, failing CI test on Py3.13/Ubuntu-lowest-direct that's been haunting this PR since its inception.

pawamoy (approved):

pawamoy:
Don't mind the 3.13-lowest failure, it's unrelated indeed and it's allowed to fail.

pawamoy:
> Concerning code style, I come from a black/ruff environment, and found no settings in your code repository. So my formatting and linting may be off.

Just saw this, sorry. There's a `make vscode` command you can use to configure VSCode for the project :slightly_smiling_face: It overrides local files.

pawamoy:
Thanks again!

pawamoy:
I refactored the logic again in https://github.com/mkdocstrings/griffe/commit/a80bd3c0cc14e5f6efc30fb804b8c7fccb319276, I think it's even more readable now.

the-13th-letter:
That's fine. I wasn't sure if I should touch the "where is the annotation looked up in?" part, so my version doesn't quite reap the same readability benefits as yours.

PR #330, opened 2024-10-26
Fix: f-string inclusion in debug msg

klieret:
Thank you for all the work on mkdocstrings! Tiny fix, I noticed that some of the messages read like

```
DEBUG   -  griffe: Iteration 1 finished, {len(resolved)} aliases resolved, still {len(unresolved)} to go
```

### DIFF
--- a/src/_griffe/loader.py
+++ b/src/_griffe/loader.py
@@ -261,8 +261,7 @@ class GriffeLoader:
                 resolved |= next_resolved
                 unresolved |= next_unresolved
             logger.debug(
-                "Iteration %s finished, {len(resolved)} aliases resolved, still {len(unresolved)} to go",
-                iteration,
+                f"Iteration {iteration} finished, {len(resolved)} aliases resolved, still {len(unresolved)} to go",
             )
         return unresolved, iteration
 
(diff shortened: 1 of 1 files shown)


src/_griffe/loader.py
```diff
@@ -261,8 +261,7 @@ def resolve_aliases(
                 resolved |= next_resolved
                 unresolved |= next_unresolved
             logger.debug(
-                "Iteration %s finished, {len(resolved)} aliases resolved, still {len(unresolved)} to go",
-                iteration,
+                f"Iteration {iteration} finished, {len(resolved)} aliases resolved, still {len(unresolved)} to go",
```
pawamoy, line 264: Lets use the %-format instead :slightly_smiling_face: Best used with logging, for perf reasons. Not more performant at all here, but for the sake of consistency.

```suggestion
                "Iteration %s finished, %s aliases resolved, still %s to go",
                iteration,
                len(resolved),
                len(unresolved),
```

pawamoy (commented):
Thanks a lot :heart:

pawamoy (approved):

Issue #352, opened 2025-01-16
bug: two issues with parsing Google-style Deprecated sections


the-13th-letter:
### Description of the bug (1 of 2)

 1. [Griffe claims to specifically support "Deprecated" sections in Google-style docstrings](https://mkdocstrings.github.io/griffe/reference/docstrings/#google-section-deprecated), but actually parses them as general admonitions.  (See the "To Reproduce" section below.)  The parsing code is there, but it isn't being triggered, because "deprecated" is not recognized as a section name: it is missing from the table of known section names.

    https://github.com/mkdocstrings/griffe/blob/a2f320fd4138a0a6697376509d625b9ec6e0797b/src/_griffe/docstrings/google.py#L48-L72

    ~I'm preparing a PR to address this.~ PR #353 addresses this issue.

### To Reproduce (1 of 2)

Using `mkdocstrings-python` 1.13.0 (with MkDocs 1.6.1 and `mkdocs-material` 9.5.49) and `griffe` 1.5.4, the following Python function

```` python
def griffe_bug() -> None:
    """bla bla bla

    Deprecated:
        Since v1.0: Insecure! Do not use!
    """
````

renders as

![(an admonition)](https://github.com/user-attachments/assets/3ac9316d-e765-4186-9780-66abe9312b81)

Changing the section heading to `DeprecatedABC` changes the rendered admonition heading to "DeprecatedABC", confirming that this is indeed parsed as an admonition.

The tests in #353 confirm this suspicion as well.

### Description of the bug (2 of 2)

If point (1) is fixed by adding the "deprecated" entry into the recognized section names table, then Griffe will assume the section contents are of the form `version: text`, and will attempt to partition the contents in this manner.  If this fails, then Griffe will issue a warning, and then **silently drop the entire section**.  This is because the parsing code, upon failing to partition the section text, returns `None` instead of a `DocstringSection` object (L711).

https://github.com/mkdocstrings/griffe/blob/a2f320fd4138a0a6697376509d625b9ec6e0797b/src/_griffe/docstrings/google.py#L706-L719

### To Reproduce (2 of 2)

This is only confirmed via the tests in #353.  No published version of Griffe exhibits this behavior yet.

### Expected behavior

Griffe should parse these sections as Deprecated sections, not as admonitions.  `mkdocstrings-python` should render them not as default admonitions, but as specific ones (perhaps similar to how griffe-deprecations renders them).

[...]

PR #353, opened 2025-01-16
fix: Parse Deprecated sections in Google-style docstrings

the-13th-letter:
This PR addresses the two issues highlighted in #352.


pawamoy:
Looks good! Was there something else you wanted to do before getting out of draft?

the-13th-letter:
Had an IRL appointment coming up, and it was evident that if something went wrong, I'd need to return to it later in the evening (i.e., now).

There's two things I still want to address:

 1. I'd like to embed the issue number in the tests for context, not just in the commit that introduces the tests.  At worst, it's redundant.  At best, it saves you a `git bisect` run.
 2. There are some quality check complaints. One of them is my fault (`src/_griffe/docstrings/google.py:713:48: Q000 Single quotes found but double quotes preferred`), and I will take care of that one, but the others are all from code I didn't touch. Do you want me to handle these as well?

pawamoy:
> 1

:+1: 

> 2

You can run `make format` for the single vs. double quotes.

No need to handle the other unrelated quality errors, I'll deal with them in another commit/PR :slightly_smiling_face:

pawamoy:
I've pushed a fix for the quality job, it should now pass. You can merge `main` or rebase on it, as you prefer.

the-13th-letter:
Rebased on 1.5.5, ready for review.  No formatting/linting errors.  There are two failing CI checks (tests), but those both appear to be failing before the tests can even run.

src/_griffe/docstrings/google.py
```diff
@@ -708,7 +709,10 @@ def _read_deprecated_section(
         version, text = text.split(":", 1)
     except ValueError:
         docstring_warning(docstring, new_offset, f"Could not parse version, text at line {offset}")
-        return None, new_offset
+        return (
+            DocstringSectionDeprecated(version="", text=text),
+            new_offset,
+        )
```
pawamoy, line 715:
```suggestion
        return DocstringSectionDeprecated(version="", text=text), new_offset
```

pawamoy (approved):
LGTM, just a formatting suggestion, not important.

pawamoy (changes requested):
Actually, lets not warn about the missing version if we allow it to be missing.

the-13th-letter:
> Actually, lets not warn about the missing version if we allow it to be missing.

Done in 60a777bf4ddd73d7f276375883ab0c56297b48c1.

6d758e299c5c3571d0db60e43a731feeb26bece7 is a rewording of 2ce96695c1842a157e4351dba37e55f2742232d4.  I don't use Angular-style commit messages, so the "format" prefix was from memory… but apparently "style" is the correct prefix.

pawamoy:
Thanks a lot!

I usually squash commits and rewrite the message (if needed) when merging, so don't bother too much with commit messages :slightly_smiling_face:

pawamoy (approved):

the-13th-letter:
Thanks as well!

PR #381, opened 2025-04-27
feat: easily construct function signature 

ISOR3X:
Added a `construct_signature` function to the `Function class`. This closely follows the way the [jinja template works for the mkdocstrings python handler](https://github.com/mkdocstrings/python/blob/e15fb741c5f347f50d918185a5343bf8d7ccc473/src/mkdocstrings_handlers/python/templates/material/_base/signature.html.jinja)

The reason behind this PR is because I want to try to integrate python api docs into a vitepress website, but I also noticed both `griffe2md` and `mkdocstrings/python/` have similar logic for building a function signature. Perhaps this code could replace that.

REF: https://github.com/mkdocstrings/griffe/discussions/376


ISOR3X:
Those errors do not seem to come from my code so I am not sure what to do. This is one of my first time contributing to open sources projects to I might need some guidance here.

pawamoy:
Thanks a lot for your PR @ISOR3X!

The errors I see in the CI logs do seem to come from your changes: you removed the `Union` import in `src/_griffe/models.py` and Ruff is complaining.

You also added spacing changes that are irrelevant to the feature here (added tabulations), could you revert these? You can run `make format` to format the code according to the project rules.

ISOR3X:
Fixed the formatting but the CI still seems to fail...

Also `make format` doesn't work for me, but I did noticed my IDE (PyCharm) automatically made a venv when opening the project, do you know something about this?

ISOR3X:
I was worried that this was forgotten, glad to see it's not!

PR #392, opened 2025-07-07
feat: Parse sphinx parameter types as Expr expressions

echoix:
This PR enhances the sphinx docstring style by parsing the type annotations as an expression, when given either with `:param a_type foo:` or a `:param foo:`+`:type foo: a_type` pair. It uses the `docstrings.utils.parse_docstring_annotation()` function, that is also used in the google docstring style parser file. 

Used on a real project, this means that the parameters types in the table can become cross references or links, event if they aren't coming from the annotations of the function signature. Before this PR, only the typing annotations in the signature render to links.

For example, on a simple mkdocs configuration for the astroid project (the one I want to use mkdocstrings with is too complex to know if problems are caused by us or mkdocstrings), this gives:
Before:
image
image

After:
image
image
image

Some tests were added with the expected expression objects, but without the `parent=` defined for `ExprName` object types, as type checking complains, see https://github.com/mkdocstrings/griffe/discussions/391. I noticed that the assertions using `as_dict()` still pass even if the parents are different, so I'm not sure if it is a bug or they are ignored.

Please make sure the asserts added at the end are sufficient, and the ones there are needed. I played around a bit when writing them. Please also make sure I correctly used the expressions as intended.

I have another feature/fix queued up related to this, when a `:param a_type foo:` has a type that contains spaces (that is not supported by the (upstream) sphinx project either, see), that ended up skipping the parameter in the generated docs, while the info was there. The test added for this helps to cover more cases related to this PR. Because of https://github.com/mkdocstrings/griffe/discussions/391, the test added fails. Since the fix is kinda of another feature, I excluded it from this PR, my first to this repo.

### DIFF
--- a/tests/test_docstrings/test_sphinx.py
+++ b/tests/test_docstrings/test_sphinx.py
@@ -275,7 +275,7 @@ def test_parse__param_field_type_field__param_section_with_type_with_parent(
         :param {SOME_NAME}: {SOME_TEXT}
         :type {SOME_NAME}: {type_}
     """
-    parent_fn = Function("func", parameters=Parameters(Parameter("foo")))
+    parent_fn = Function("func", parameters=Parameters(Parameter(SOME_NAME)))
     sections, _ = parse_sphinx(docstring, parent=parent_fn)
     assert len(sections) == 2
     assert sections[1].kind is DocstringSectionKind.parameters
@@ -324,7 +324,7 @@ def test_parse__param_field_type_field_first__param_section_with_type_with_paren
         :type {SOME_NAME}: str
         :param {SOME_NAME}: {SOME_TEXT}
     """
-    parent_fn = Function("func", parameters=Parameters(Parameter("foo")))
+    parent_fn = Function("func", parameters=Parameters(Parameter(SOME_NAME)))
     sections, _ = parse_sphinx(docstring, parent=parent_fn)
     assert len(sections) == 2
     assert sections[1].kind is DocstringSectionKind.parameters
@@ -394,7 +394,7 @@ def test_parse__param_field_type_field_or_none__param_section_with_optional_with
         :type {SOME_NAME}: {union}
     """
 
-    parent_fn = Function("func", parameters=Parameters(Parameter("foo")))
+    parent_fn = Function("func", parameters=Parameters(Parameter(SOME_NAME)))
     sections, _ = parse_sphinx(docstring, parent=parent_fn)
     assert len(sections) == 2
     assert sections[1].kind is DocstringSectionKind.parameters


pawamoy (changes requested):
Looking good, thanks a lot! Just a few nitpicks.

tests/test_docstrings/test_sphinx.py
```diff
@@ -231,28 +234,62 @@ def test_parse__param_field_docs_type__param_section_with_type(parse_sphinx: Par
...
+        parse_sphinx: Fixture parser.
+        type_: The type to use in the type directive.
+    """
+    docstring = f"""
+        Docstring with line continuation.
+
+        :param {SOME_NAME}: {SOME_TEXT}
+        :type {SOME_NAME}: {type_}
+    """
+    parent_fn = Function("func", parameters=Parameters(Parameter("foo")))
```
pawamoy, line 278:
```suggestion
    parent_fn = Function("func", parameters=Parameters(Parameter(SOME_NAME)))
```

tests/test_docstrings/test_sphinx.py
```diff
@@ -275,6 +312,33 @@ def test_parse__param_field_type_field_first__param_section_with_type(parse_sphi
...
+    Parameters:
+        parse_sphinx: Fixture parser.
+    """
+    docstring = f"""
+        Docstring with line continuation.
+
+        :type {SOME_NAME}: str
+        :param {SOME_NAME}: {SOME_TEXT}
+    """
+    parent_fn = Function("func", parameters=Parameters(Parameter("foo")))
```
pawamoy, line 327:
```suggestion
    parent_fn = Function("func", parameters=Parameters(Parameter(SOME_NAME)))
```

tests/test_docstrings/test_sphinx.py
```diff
@@ -302,6 +366,47 @@ def test_parse__param_field_type_field_or_none__param_section_with_optional(
...
+        expected_annotation: The expected annotation as an expression
+    """
+    docstring = f"""
+        Docstring with line continuation.
+
+        :param {SOME_NAME}: {SOME_TEXT}
+        :type {SOME_NAME}: {union}
+    """
+
+    parent_fn = Function("func", parameters=Parameters(Parameter("foo")))
```
pawamoy, line 397:
```suggestion
    parent_fn = Function("func", parameters=Parameters(Parameter(SOME_NAME)))
```

pawamoy (approved):
Just a few more, let me batch-commit them and we're good to merge :slightly_smiling_face:

echoix:
> Just a few more, let me batch-commit them and we're good to merge 🙂

There’s other tests (not in the diff) that could have been changed, i limited myself to the scope of the PR

pawamoy:
Yep that's good :+1:

PR #430, opened 2025-11-17
Change build system to hatchling

johnslavik:
Seems like it outputs the same thing.

### DIFF
--- a/src/griffe/_internal/agents/inspector.py
+++ b/src/griffe/_internal/agents/inspector.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 
 import ast
 import functools
-import sys
 import types
 import typing
 from inspect import Parameter as SignatureParameter
--- a/src/griffe/_internal/expressions.py
+++ b/src/griffe/_internal/expressions.py
@@ -7,7 +7,6 @@
 from __future__ import annotations
 
 import ast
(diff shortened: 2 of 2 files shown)


johnslavik:
Did this simple test setup:

```bash
#!/usr/bin/env bash

set -HCETeuxo pipefail

TEST_TAG="v9.9.99"

rm -rf {before,after}
mkdir -p {before,after}/{sdist,wheel}

git checkout main
git tag -fsam 'Test' "$TEST_TAG"
uv build --{sdist,wheel} -o before

git checkout b/use-hatchling-build-system
git tag -fsam 'Test' "$TEST_TAG"
uv build --{sdist,wheel} -o after

tar -xzf before/*.gz -C before/sdist
tar -xzf after/*.gz -C after/sdist

unzip before/*.whl -d before/wheel
unzip after/*.whl -d after/wheel

git tag -d "$TEST_TAG"

for dir in {before,after}/{sdist,wheel}; do
    fname="${dir//\//.}"
    du -Sh "$dir" | sort -h >| "du.$fname"
    tree "$dir" >| "tree.$fname"
done
```

Comparing `du.before.sdist` and `du.after.sdist` caught that the sdist is missing the `src` directory, gonna fix it

PR #431, opened 2025-11-18
 change: Reorganize as `griffe` and `griffelib` packages with uv workspaces

johnslavik:
I've started with this incomplete migration script (make sure to not have a dirty env, I personally run `git clean -dxfe migrate.py -e .vscode -e .venv -e .venvs -e uv.lock && git reset --hard main`): https://gist.github.com/johnslavik/76d556a427033c4de2d236d67aa05b8f

Will carry out the remaining details manually.


johnslavik:
It's good now

johnslavik:
Given how workspaces and packaging work, we'll do it a little different:

Put `griffelib` in the root (code under `src/`) and `griffe` in `packages/griffe/`.

PR #432, opened 2025-11-19
Clarify docstring Warnings section usage

AntoineD:
I was struggling on trying to use `Warnings` in docstrings in google format, I was even about to create a fix when I remembered that I already went [this path](https://github.com/mkdocstrings/griffe/pull/209) a couple years ago! :laughing:


pawamoy:
No worries, thank you for your patience :slightly_smiling_face: 

Yes, the tip would be enough IMO. Maybe with a "See Admonitions" link.

PR #433, opened 2025-11-29
Prepare for 2.0 (finalize deprecations)

johnslavik:
The #431 migration will be way less pain if we cleanup old deprecations first.


src/griffe/_internal/extensions/base.py
```diff
@@ -647,7 +611,9 @@ def load_extensions(*exts: LoadableExtensionType) -> Extensions:
 
     # TODO: Deprecate and remove at some point?
     # Always add our built-in dataclasses extension.
-    from griffe._internal.extensions.dataclasses import DataclassesExtension  # noqa: PLC0415
+    from griffe._internal.extensions.dataclasses import (
+        DataclassesExtension,
+    )
```
pawamoy: nit: Can you put that back on one line?

src/griffe/_internal/extensions/base.py
```diff
@@ -25,43 +22,18 @@
...
-class Extension(metaclass=_ExtensionMetaclass):
+    from griffe._internal.models import (
+        Alias,
+        Attribute,
+        Class,
+        Function,
+        Module,
+        Object,
+        TypeAlias,
+    )
```
pawamoy: nit: Same here, formatted with default line-length instead of 120?

pawamoy (commented):
Yes, that's a good idea. Since we're going to bump the major version for the griffe/griffelib split, lets remove deprecated code at the same time :+1:

src/griffe/_internal/extensions/base.py
johnslavik:
```suggestion
    from griffe._internal.extensions.dataclasses import DataclassesExtension # noqa: PLC0415
```

src/griffe/_internal/extensions/base.py
johnslavik:
```suggestion
    from griffe._internal.models import Alias, Attribute, Class, Function, Module, Object, TypeAlias
```

pawamoy (approved):
Looks good :+1:

PR #434, opened 2025-11-30
change: Split to `griffe` + `griffecli` + `griffelib` uv workspace

### LINKED ISSUE #408: change: Split `griffe` to `griffe` and `griffelib` packages
We'd like to make this repository a monorepo of 2 packages (and PyPI projects) managed with uv workspaces:
- `griffelib`, containing all library logic
- `griffe`, a thin package that can be run as script which wraps `griffelib` with a CLI

This originates from discussions in #406.
However, when working through this, I realized that it's a huge enough migration that quite a few things could be worth being explicitly mentioned and considered.

## Our goals
- All variants of `pip install griffe` install a CLI application and the library. `uvx griffe`, `pipx run griffe` or `pip-run griffe` naturally run the griffe CLI.
- Public API imports from griffe, such as `from griffe import ...` imports, still work, and `griffe/__init__.py` is just a shim to all public members of the `griffelib` library.

## Things to consider
- I'm not planning to add shims to new `griffe/_internal` modules -- they're called `_internal` for a reason, and btw @pawamoy that's awesome we have it this way. Just note that this will break a few packages -- but [few enough](https://github.com/search?q=%22from+griffe._internal%22+-org%3Amkdocstrings+-org%3Apawamoy-insiders+-repo%3AAA-Turner%2Ftop-pypi-sdists-2000+-repo%3Aechoix%2Fgriffe+-path%3A%2F.*griffe.*%2F&type=code) to just go to their repos and kindly inform them that they'll need to migrate to `griffelib` if they want to maintain their unstable imports. Perhaps they don't even have to use internal imports? I haven't inspected that deeply.
- Needs decision: Should we add a warning to all `from griffe ... import` library imports, and recommend to import library things from `griffelib` instead? I'd say no, because maintaining a `from griffe import *` is low-cost, so I wouldn't care but perhaps you'd like to drop that. Do most people use `griffe` as an app, not a library? @pawamoy 

## Migration plan
- [ ] Introduce uv workspaces
- [ ] Introduce uv dynamic versioning for syncing `griffe` and `griffelib` releases to be exact same version. As discussed privately, we'd like the two packages to be released with 1:1 versioning, even if one of them doesn't change, as we find a 1:1 relationship easier to reason about
- [ ] Migrate tests and CI to new monorepo layout
- [ ] Migrate documentation
  - [ ] Present `griffelib` and `griffe` separately in our documentation
  - [ ] Migrate crossrefs in the docstrings -- thankfully we have this nice tool called `griffe` that will help validate that we did it correctly
  - [ ] Important: make sure the old crossref links redirect to the new crossref targets
- [ ] Release a new version `griffelib` and `griffe`

Not sure yet how to partition that into PRs. I'll come up with whatever makes sense.

johnslavik:
Continuation of #431, based on #433, closes #408
Migration script: https://gist.github.com/johnslavik/76d556a427033c4de2d236d67aa05b8f


johnslavik:
Let's try Copilot here

johnslavik:
@copilot We've migrated griffe to 3 packages exposed by `griffe` package in a backward-compatible way, all bound by uv dynamic versioning that ensures compatibility. `griffecli` is supposed to only implement and expose CLI (which needs proper configuration in pyproject.toml) and depend on `griffelib=={{version}}`. `griffelib` is supposed depend on nothing and implement the necessary logic. `griffe` is supposed to be a backward-compatible package depending on both `griffelib` and `griffecli`, with all the previous packaging and runtime behaviors in regular use as before the migration.

Fix any instances of name errors (import griffe, but then griffelib is used, etc.), inconsistencies caused by the migration, and any references broken by the migration script. Make sure that the griffecli's init module is correct. Do complete whatever the migration script missed, making sure the migration is fully complete and backward-compatible. Update the documentation to reflect new changes, if necessary. Document all changes that you've made.

johnslavik:
@copilot Add missing configuration from pyproject.toml (check CI failures) and make sure that we promote the old way of running things (`griffe check`, not `griffecli check`; `python -m griffe`, not `python -m griffecli`) -- but document that it is possible to use `python -m griffecli` instead of `python -m griffe` where it makes sense.

johnslavik:
@copilot Look at https://github.com/mkdocstrings/griffe/actions/runs/19806731977/job/56741990187?pr=434. Fix all incorrect refs

packages/griffecli/src/griffecli/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+The CLI (Command Line Interface) for the griffe library.
+This package provides command-line tools for interacting with griffe.
+
+## CLI entrypoints
+
+- [`griffecli.main`][]: Run the main program.
+- [`griffecli.check`][]: Check for API breaking changes in two versions of the same package.
+- [`griffecli.dump`][]: Load packages data and dump it as JSON.
+- [`griffecli.get_parser`][]: Get the argument parser for the CLI.
```
pawamoy: I see a mix of cross-ref, some using `griffe.*`, some using `griffecli.*` or `griffelib.*`, we should make them consistent.

packages/griffelib/src/griffelib/_internal/diff.py
```diff
@@ -150,7 +150,7 @@ def _relative_package_filepath(self) -> Path:
     def _location(self) -> Path:
         # Absolute file path probably means temporary worktree.
         # We use our worktree prefix to remove some components
-        # of the path on the left (`/tmp/griffe-worktree-*/griffe_*/repo`).
+        # of the path on the left (`/tmp/griffelib-worktree-*/griffe_*/repo`).
```
pawamoy: I think we could keep these foders named "griffe" rather than "griffelib", as the distinction is not important in this case.

packages/griffelib/src/griffelib/_internal/loader.py
```diff
@@ -983,7 +983,7 @@ def load_pypi(
 
     import platformdirs  # noqa: PLC0415
 
-    pypi_cache_dir = Path(platformdirs.user_cache_dir("griffe"))
+    pypi_cache_dir = Path(platformdirs.user_cache_dir("griffelib"))
```
pawamoy: Same here, we could keep using `griffe`.

scripts/gen_griffe_json.py
```diff
@@ -5,5 +5,5 @@
 python_handler = mkdocs_gen_files.config.plugins["mkdocstrings"].get_handler("python")
 data = python_handler.collect("griffe", options=python_handler.get_options({}))
 
-with mkdocs_gen_files.open("griffe.json", "w") as fd:
+with mkdocs_gen_files.open("griffelib.json", "w") as fd:
```
pawamoy: We'll have to update this somehow, to contain griffe, griffecli and griffelib.

packages/griffelib/src/griffelib/_internal/diff.py
johnslavik: Yup

johnslavik:
I'm going over the whole pipeline (docs, usage, etc) now, addressing these and other issues on the way.

johnslavik:
I'm thinking we shouldn't add `griffelib` as an importable itself. Just reuse `griffe` as the namespace.

packages/griffelib/src/griffe/_internal/finder.py
```diff
@@ -483,7 +483,7 @@ def _handle_editable_module(path: Path) -> list[_SP]:
         return [_SP(new_path)]
     if _match_pattern(path.name, _editable_setuptools_patterns):
         # Support for how 'setuptools' writes these files:
-        # example line: `MAPPING = {'griffe': '/media/data/dev/griffe/src/griffe', 'briffe': '/media/data/dev/griffe/src/briffe'}`.
+        # example line: `MAPPING = {'griffe': '/media/data/dev/griffe/packages/griffelib/src/griffe', 'briffe': '/media/data/dev/griffe/src/briffe'}`.
```
pawamoy:
```suggestion
        # example line: `MAPPING = {'pkg': '/data/dev/pkg/src/pkg', 'pkg2': '/data/dev/pkg/src/pkg2'}`.
```

pawamoy (changes requested):

packages/griffelib/src/griffe/_internal/finder.py
```diff
@@ -508,7 +508,7 @@ def _handle_editable_module(path: Path) -> list[_SP]:
                 and node.value.func.id == "install"
                 and isinstance(node.value.args[1], ast.Constant)
             ):
-                build_path = Path(node.value.args[1].value, "src")  # type: ignore[arg-type]
+                build_path = Path(node.value.args[1].value, "packages/griffelib/src")  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
johnslavik: Working on the fix. Thank you for pointing me to the right direction, I wasn't thoughtful on this one!

packages/griffelib/src/griffe/_internal/debug.py
```diff
@@ -84,7 +84,7 @@ def _get_debug_info() -> _Environment:
         Environment information.
     """
     py_name, py_version = _interpreter_name_version()
-    packages = ["griffe"]
+    packages = ["griffelib"]
```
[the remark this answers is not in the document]
johnslavik: Reading with comprehension :exploding_head: 

Thanks :)

packages/griffelib/src/griffe/_internal/finder.py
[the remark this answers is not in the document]
pawamoy: Thanks, of course! Don't hesitate to use something else than "griffe" in tests if that makes it easier / less confusing / less error-prone :slightly_smiling_face:

johnslavik:
We are suddenly hit with failing UP007 checks (PEP 604 type annotation syntax). I added it to ignored rules in this PR because it is out of scope. I'll fix it in a separate PR.

pawamoy (approved):
OK, LGTM! Thanks for all the hard work on this! 😄

PR #440, opened 2025-12-05
Add support for PEP 750 template strings

1borgy:
### For reviewers

- [X] I did not use AI

### Description of the change

Add expressions for `ast.TemplateStr` and `ast.Interpolation`, which were added in Python 3.14. Round-tripping (e.g. `tests/test_expressions.py::test_expressions`) does not work with the `conversion` or `format_spec` fields, though this is consistent with the current implementation of `ExprJoinedStr` and `ExprFormatted`.

### Relevant resources

- https://peps.python.org/pep-0750/
- https://docs.python.org/3/library/ast.html#ast.TemplateStr
- https://docs.python.org/3/library/ast.html#ast.Interpolation


1borgy:
Looks like most of CI is passing, bar `config/test_api.py::test_api_matches_inventory` for python<3.14. The suggestion of `make run mkdocs build` did not generate any changes to `site/objects.inv` for me. For the sake of not overusing GHA runner resources, I'll wait for a maintainer's blessing before investigating further.

tests/test_nodes.py
```diff
@@ -51,6 +52,7 @@
     "call(something=something)",
     # Strings.
     "f'a {round(key, 2)} {z}'",
+    *(["t'a {round(key, 2)} {z}'"] if sys.version_info >= (3, 14) else []),
```
1borgy: could also try something like
```suggestion
    pytest.param("t'a {round(key, 2)} {z}'", marks=pytest.mark.skipif(sys.version_info < (3, 14))),
```

pawamoy (commented):
Thanks @1borgy, super clean! Just a few comments.

pawamoy:
Yep, failures on 3.15 are fine, I'll investigate them later :slightly_smiling_face: Let me know when you think this is ready to merge!

PR #455, opened 2026-05-16
fix: Render f-strings and t-strings with correct quote delimiters

### LINKED ISSUE #444: bug: ExprJoinedStr and ExprTemplateStr don't always produce correct code
### Description of the bug

These two classes produce code by hardcoding single quotes at the beginning and end of the strings. This can produce invalid code if the string contains additional single quotes.

### To Reproduce

TODO

### Full traceback

/

### Expected behavior

We should produce valid code, always.

### Environment information

```bash
griffe --debug-info  # | xclip -selection clipboard
```

- __System__: Linux-6.18.7-arch1-1-x86_64-with-glibc2.42
- __Python__: cpython 3.14.2 (/run/media/pawamoy/Data/dev/griffe/.venv/bin/python3)
- __Environment variables__:
- __Installed packages__:
  - `griffelib` v1.15.1.dev12+g162c31c2
  - `griffecli` v1.15.1.dev12+g162c31c2
  - `griffe` v1.15.1.dev12+g162c31c2

### Additional context

Spotted in https://github.com/mkdocstrings/griffe/pull/440

johnslavik:
Closes #444

Still in progress (I'm validating the implementation), but that's more or less what I'd like to land.

`ExprJoinedStr.iterate()` and `ExprTemplateStr.iterate()` previously hardcoded a single-quote delimiter, producing invalid Python source when the literal parts of the f-string contained apostrophes.

Port ast.unparse's visit_JoinedStr quote-selection algorithm: pre-render all parts (literals get brace-doubled and control-char-escaped, expression parts get flattened to string), then pick the first quote type from ("'", '"', "'''", '"""') that doesn't conflict with any part.

Also guard _build_constant's parse_strings path with `not in_formatted_str` so string literals inside f-string expressions ({...}) are never mistakenly compiled as type annotations.

### For reviewers

- [ ] I did not use AI

### DIFF
--- a/packages/griffelib/src/griffe/_internal/expressions.py
+++ b/packages/griffelib/src/griffe/_internal/expressions.py
@@ -563,6 +563,47 @@ class ExprInterpolation(Expr):
         yield "}"
 
 
+_FSTRING_ALL_QUOTES = ("'", '"', "'''", '"""')
+_FSTRING_MULTI_QUOTES = ('"""', "'''")
+
+
+def _fstring_choose_quote(values: Sequence[str | Expr]) -> tuple[str, list[str]]:
+    # Follows ast.unparse's visit_JoinedStr quote-selection algorithm.
+    # Pre-render all parts: literals get brace-doubled, expressions get flattened.
+    fstring_parts: list[tuple[str, bool]] = []
+    for value in values:
+        if isinstance(value, str):
+            fstring_parts.append((value.replace("{", "{{").replace("}", "}}"), True))
+        else:
+            rendered = "".join(
+                p if isinstance(p, str) else p.name
+                for p in value.iterate(flat=True)
+            )
+            fstring_parts.append((rendered, False))
+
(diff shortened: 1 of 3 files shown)


pawamoy:
Thanks a lot @johnslavik, looks great :slightly_smiling_face: 

About flattening expressions to a string: it's a bit sad that we loose the ability to iterate on each expression part :thinking: I'm trying to see if there's a way to retain this ability while also fixing quoting.

- keep old, broken behavior in addition to new one, thanks to parameter: meh
- somehow reparse fixed string to expression, properly keeping surrounding quotes info?
- something else?

johnslavik:
> ability to iterate on each expression part 🤔

Please help me understand: why is it preferred to retain it?

pawamoy:
Well you raised a good point about being able to cross-references inside of formatted strings :thinking: And also: breaking change. But anyway, I don't think we were going to start cross-referencing stuff in formatted strings, not with this system anyway. So, you can consider this done :+1: I'll take a bit more time to review again before merging :relaxed:

PR #465, opened 2026-06-19
Parse indented numpy docstring examples

gselzer:
### For reviewers

- [ ] I did not use AI
- [X] I used AI and thoroughly reviewed every code/docs change

I've never dug into mkdocstrings before, so the LLM was useful for pinpointing the issue and suggesting a quick fix. Afterwards I slimmed down a rather verbose test 😅

### Description of the change

This PR is designed to enable the parsing of numpy-style code examples that are nonuniformly indented. Sometimes authors (read: me) might want to indent a code example to highlight it amidst the larger section - but griffe doesn't pick those up.

This PR also includes a test that fails without the changes.

### Relevant resources

This issue was spotted in https://github.com/pyapp-kit/scenex/issues/89

### DIFF
--- a/packages/griffelib/src/griffe/_internal/docstrings/numpy.py
+++ b/packages/griffelib/src/griffe/_internal/docstrings/numpy.py
@@ -816,7 +816,7 @@ def _read_examples_section(
                 current_text.append(line)
 
         elif in_code_example:
-            line = line.lstrip()
+            line = line.lstrip()  # noqa: PLW2901
             if trim_doctest_flags:
                 line = _RE_DOCTEST_FLAGS.sub("", line)  # noqa: PLW2901
                 line = _RE_DOCTEST_BLANKLINE.sub("", line)  # noqa: PLW2901
--- a/packages/griffelib/src/griffe/_internal/docstrings/numpy.py
+++ b/packages/griffelib/src/griffe/_internal/docstrings/numpy.py
@@ -836,7 +836,7 @@ def _read_examples_section(
             in_code_example = True
 
             if trim_doctest_flags:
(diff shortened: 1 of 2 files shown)


pawamoy:
Thanks for the PR @gselzer. Have you considered simply *not* indenting your `>>>` blocks and use https://github.com/pawamoy/markdown-pycon to enable proper highlighting once rendered through MkDocs/Zensical?

Also the rendering issue in https://github.com/pyapp-kit/scenex/issues/89 might simply be due to the lack of a blank line between the markup and the `>>>` block.

gselzer:
> Have you considered simply _not_ indenting your `>>>` blocks

Yes, that's what pyapp-kit/scenex#93 does, however it is clear to me now that I am prone to doing this, and in the case others are as well I figured a fix upstream couldn't hurt 😅 

> and use https://github.com/pawamoy/markdown-pycon to enable proper highlighting once rendered through MkDocs/Zensical?

Umm, maybe, how is that different from the examples within the docstrings are being rendered now without pycon?

image

> Also the rendering issue in [pyapp-kit/scenex#89](https://github.com/pyapp-kit/scenex/issues/89) might simply be due to the lack of a blank line between the markup and the `>>>` block.

Based on a quick change and a local Zensical build that doesn't seem to be part of the issue 🙁

pawamoy:
Ah right the markdown-pycon extension makes it possible to write `>>>` *anywhere*, not just in `Examples` sections understood by Griffe.

Since de-denting such blocks in Examples sections works as expected, I'm gonna close this PR. I'll reconsider if more people run into this :slightly_smiling_face: 

Thanks again for your contribution!

PR #466, opened 2026-06-20
fix: Keep quotes on string members of unresolved Literal annotations

gaoflow:
### For reviewers

- [ ] I did not use AI

### Description of the change

A `Literal[...]` whose name could not be resolved (for example used without `from typing import Literal`) keeps its bare name `"Literal"` as its canonical path. The guard in `_build_subscript` that enables `literal_strings` only matched the resolved spellings `typing.Literal` / `typing_extensions.Literal`, so for an unresolved `Literal` the string members were treated as forward references and re-parsed:

```python
# unresolved Literal (no import)
x: Literal['a-b']      # -> Literal[a - b]   (parsed as a subtraction!)
y: Literal['a', 'b']   # -> Literal[a, b]    (names, not strings)
```

A resolved `Literal` keeps the quotes (`Literal['a-b']`), and value-position strings always keep theirs, so the two paths were inconsistent.

This adds the bare `"Literal"` spelling to that set. A genuinely user-defined `Literal` resolves to a dotted path (e.g. `mod.Literal`), so it is not over-matched, and forward references elsewhere (e.g. `list['Foo']` → `list[Foo]`) are unaffected.

Verified:
- `Literal['a-b']`, `Literal['a', 'b']`, `Literal['hello world']` (unresolved) now round-trip with quotes intact.
- Regression cases hold: imported `Literal`, `typing.Literal[...]`, aliased import, int literals, and `list['Foo']` still resolving its forward reference.
- Added a parametrized regression test in `tests/test_expressions.py` (it fails before the change, passes after). `test_expressions.py`: 146 passed. `ruff check` clean (zero new warnings).

### Relevant resources

-
-
-

### DIFF
--- a/packages/griffelib/src/griffe/_internal/expressions.py
+++ b/packages/griffelib/src/griffe/_internal/expressions.py
@@ -1324,6 +1324,10 @@ def _build_subscript(
         if isinstance(left, (ExprAttribute, ExprName)) and left.canonical_path in {
             "typing.Literal",
             "typing_extensions.Literal",
+            # A `Literal` that could not be resolved (e.g. used without an
+            # import) keeps its bare name as canonical path. Its string members
+            # are literal strings, not forward references, so keep their quotes.
+            "Literal",
         }:
             literal_strings = True
         slice_expr = _build(
--- a/packages/griffelib/tests/test_expressions.py
+++ b/packages/griffelib/tests/test_expressions.py
@@ -61,6 +61,26 @@ def test_full_expressions(annotation: str) -> None:
         assert str(module["x"].annotation) == annotation
 
 
+@pytest.mark.parametrize(
+    "annotation",
+    [
+        "Literal['a-b']",
+        "Literal['a', 'b']",
+        "Literal['hello world']",
+    ],
+)
+def test_unresolved_literal_keeps_string_quotes(annotation: str) -> None:
+    """Assert string members of an unresolved `Literal` keep their quotes.
+
+    An unresolved `Literal` (used without an import) must still treat its
+    members as literal strings rather than forward references, otherwise
+    `Literal['a-b']` is mis-parsed as `Literal[a - b]`.
+    """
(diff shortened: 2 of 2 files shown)


pawamoy:
Thanks for the PR @gaoflow.

Can you tell me why this is useful? In what case would someone use `Literal` without importing it?

binggao1230:
Thanks. I agree that valid code should normally import it; I am not trying to encourage using `Literal` without an import.

The useful part here is mostly defensive/lossless parsing for source that Griffe can parse but cannot resolve completely. For example, with `from __future__ import annotations`, a module can still be imported even if an annotation name is missing or guarded differently, and documentation tooling may see partial/in-progress code before a type checker catches it. In that situation I think Griffe should preserve the source shape rather than rewrite string literal members into different expressions.

The current behavior changes meaning quite a bit:

```python
x: Literal['a-b']
```

is rendered as `Literal[a - b]`, and `Literal['a', 'b']` as `Literal[a, b]`. Even if the annotation is semantically unresolved, the string members are still literal strings in the source, not forward references.

The change is intentionally narrow: it only handles the bare unresolved canonical path `Literal`, while a real user-defined `Literal` resolves to a dotted path such as `mod.Literal`, so it is not matched.

pawamoy:
Thanks. I hear the argument about partial code (code being edited while tools are running). However I think Griffe should simply *not crash*, rather than *guess* what the user's intent is. So we're already covered since Griffe doesn't crash on this partial code.

> In that situation I think Griffe should preserve the source shape

I also agree with that. However the fix here makes a special case of `Literal` rather than apply the "preserve source shape" suggestion generally. I'm not sure yet how we could preserve source shape generally, and when (inside subscripts of variables that aren't defined?).

For these two reasons, and because there doesn't seem to be an actual, real-world example of why this would be useful in practice, I'm gonna close this :slightly_smiling_face: I hope you can understand! Thanks again for contributing to the project :pray:

PR #468, opened 2026-06-22
docs: Fix google-style docstrings in `duties.py`

mushitoriami:
### For reviewers

- [x] I did not use AI

### Description of the change

In the docstrings in `duties.py`, there are some `Example:` occurrences, but these should be `Examples:` (https://mkdocstrings.github.io/griffe/reference/docstrings/#google-section-examples). This issue causes broken document format (https://mkdocstrings.github.io/griffe/guide/contributors/commands/#check-quality, https://mkdocstrings.github.io/griffe/guide/contributors/commands/#check-types ).

### DIFF
--- a/duties.py
+++ b/duties.py
@@ -128,7 +128,7 @@ def check_quality(ctx: Context) -> None:
 
     ...where ID is the identifier of the rule you want to ignore for this line.
 
-    Examples:
+    Example:
         ```python title="src/your_package/module.py"
         import subprocess
         ```
@@ -220,7 +220,7 @@ def check_types(ctx: Context) -> None:
 
     ...where ID is the name of the warning.
 
-    Examples:
+    Example:
         ```python title="src/your_package/module.py"
         result = data_dict.get(key, None).value
         ```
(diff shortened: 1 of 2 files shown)


pawamoy:
Thanks @mushitoriami! I don't understand though, the Example section should definitely work, it should create an admonition. If I load the code with Griffe and parse the docstring, the section is correctly parsed as an admonition :upside_down_face: So I don't understand yet why it's not rendered as such.

pawamoy:
Ah, right, I'm dog-fooding the `auto` docstrings parser, and in this case no style is inferred (so no Google-style admonition parsing). The fix is to either fix the function that infers the style, or set `google` in mkdocs.yml. Would you like to update the PR?

mushitoriami:
@pawamoy Thanks for the comment. I've updated to set the style option.

mushitoriami:
It looks like the CI failure is caused by `ty==0.0.52`, which was released today.

pawamoy:
Thanks! 
Yeah I'll fix CI elsewhere.

pawamoy:
Actually, lets rather add some patterns to the function that infers style. Let me do that in another PR. Thanks again for the report!

PR #474, opened 2026-06-25
fix: render empty-tuple subscript annotation as tuple[()] not tuple[]

gaoflow:
### For reviewers

- [ ] I did not use AI

### Description of the change

Fix rendering for empty-tuple subscript annotations so `tuple[()]` and `Tuple[()]` stringify as valid Python instead of `tuple[]` / `Tuple[]`.

`tuple[()]` is a valid annotation for a zero-element tuple. Griffe parses it as an `ExprSubscript` whose slice is an implicit, empty `ExprTuple`. Because `ExprTuple.iterate()` suppressed parentheses when `implicit=True` and yielded no elements for an empty tuple, the outer subscript rendered empty brackets.

An empty tuple has no omitted source form; it must be written as `()`. This adds a focused early return for the empty implicit tuple case and covers both builtin `tuple` and `typing.Tuple` with `test_empty_tuple_annotation_str`.

Validation:

- Added `test_empty_tuple_annotation_str`.
- All 977 tests pass locally.

### Relevant resources

- No linked issue.

### DIFF
--- a/packages/griffelib/src/griffe/_internal/expressions.py
+++ b/packages/griffelib/src/griffe/_internal/expressions.py
@@ -1003,6 +1003,14 @@ class ExprTuple(Expr):
     """Whether the tuple is implicit (e.g. without parentheses in a subscript's slice)."""
 
     def iterate(self, *, flat: bool = True) -> Iterator[str | Expr]:
+        if self.implicit and not self.elements:
+            # An empty tuple is always written as `()` and cannot be implicit.
+            # This arises in annotations like `tuple[()]`, where the AST represents
+            # the subscript slice as an empty Tuple node, but the parentheses must
+            # be preserved to produce valid Python (`tuple[]` is a SyntaxError).
+            yield "("
+            yield ")"
+            return
         if not self.implicit:
             yield "("
         yield from _join(self.elements, ", ", flat=flat)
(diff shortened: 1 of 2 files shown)


packages/griffelib/src/griffe/_internal/expressions.py
```diff
@@ -1003,6 +1003,14 @@ class ExprTuple(Expr):
     """Whether the tuple is implicit (e.g. without parentheses in a subscript's slice)."""
 
     def iterate(self, *, flat: bool = True) -> Iterator[str | Expr]:
+        if self.implicit and not self.elements:
```
pawamoy: > An empty tuple is always written as `()`

So we don't need to check `self.implicit`, right?

pawamoy (commented):
Thanks.

Please follow our PR template.

binggao1230:
Updated the PR description to follow the template, including the AI disclosure checkbox.

mushitoriami:
This issue doesn't seem to be in the Expr-to-str conversion, but in the AST-to-Expr conversion. The problem is that the `ExprTuple` corresponding to `()` should have `implicit=False`, but it doesn't.

pawamoy:
@mushitoriami good point. Maybe we can fix this in both locations (when building the expression and when converting it to a string), to be defensive.

pawamoy (approved):
I addressed both points. Thanks for the contribution!

PR #478, opened 2026-07-16
fix: Make stringified expressions valid and faithful Python

gaoflow:
### For reviewers

- [ ] I did not use AI

### Description of the change

`str(Expr)` could emit invalid or semantically different Python for lambda parameter markers, formatted strings, generator expressions, integer-literal attributes, slices, and precedence-sensitive contexts.

This fixes the shared rendering and building paths and adds a parse → Expr → str → parse matrix that checks AST equivalence across expression shapes and contexts. The full griffelib suite passes on Python 3.12; the focused expression and node suites pass on Python 3.10 and 3.14.

### Relevant resources

- #474
- https://docs.python.org/3/reference/expressions.html
- https://docs.python.org/3/library/ast.html#ast.unparse

### DIFF
--- a/packages/griffelib/src/griffe/_internal/expressions.py
+++ b/packages/griffelib/src/griffe/_internal/expressions.py
@@ -111,6 +111,7 @@ def _join(
     joint: str | Expr,
     *,
     flat: bool = True,
+    outer_precedence: _OperatorPrecedence = _OperatorPrecedence.NONE,
 ) -> Iterator[str | Expr]:
     """Apply a separator between elements.
 
@@ -119,14 +120,14 @@ def _join(
     """
     it = iter(elements)
     try:
-        # Since we are in a sequence, don't parenthesize items.
+        # Since we are in a sequence, don't parenthesize items (unless told otherwise).
         # Avoids [a + b, c + d] being serialized as [(a + b), (c + d)]
-        yield from _yield(next(it), flat=flat, outer_precedence=_OperatorPrecedence.NONE)
+        yield from _yield(next(it), flat=flat, outer_precedence=outer_precedence)
     except StopIteration:
         return
     for element in it:
-        yield from _yield(joint, flat=flat, outer_precedence=_OperatorPrecedence.NONE)
-        yield from _yield(element, flat=flat, outer_precedence=_OperatorPrecedence.NONE)
+        yield from _yield(joint, flat=flat, outer_precedence=outer_precedence)
+        yield from _yield(element, flat=flat, outer_precedence=outer_precedence)
 
 
 def _field_as_dict(
@@ -277,7 +278,12 @@ class ExprAttribute(Expr):
 
     def iterate(self, *, flat: bool = True) -> Iterator[str | Expr]:
         precedence = _get_precedence(self)
-        yield from _yield(self.values[0], flat=flat, outer_precedence=precedence, is_left=True)
+        first = self.values[0]
+        if isinstance(first, str) and first.isdigit():
+            # Integer literals need parentheses: `1.bit_length()` is a syntax error.
+            yield f"({first})"
+        else:
+            yield from _yield(first, flat=flat, outer_precedence=precedence, is_left=True)
         for value in self.values[1:]:
             yield "."
             yield from _yield(value, flat=flat, outer_precedence=precedence)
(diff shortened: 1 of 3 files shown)


pawamoy:
Thanks for your contribution. Again, please follow our PR template.

packages/griffelib/src/griffe/_internal/expressions.py
```diff
@@ -277,7 +278,12 @@ class ExprAttribute(Expr):
 
     def iterate(self, *, flat: bool = True) -> Iterator[str | Expr]:
         precedence = _get_precedence(self)
-        yield from _yield(self.values[0], flat=flat, outer_precedence=precedence, is_left=True)
+        first = self.values[0]
+        if isinstance(first, str) and first.isdigit():
+            # Integer literals need parentheses: `1.bit_length()` is a syntax error.
+            yield f"({first})"
```
pawamoy, line 284: `iterate` is used to iterate on elements, so we should yield each elements separately instead of building a string:

```suggestion
            yield "("
            yield first
            yield ")"
```

Unless you think we should consider `(1)` as a single element. I'd be OK with that too.

UPDATE: yeah, lets keep it as a single element, it makes no difference when iterating since `first` is a string.

packages/griffelib/src/griffe/_internal/expressions.py
```diff
@@ -498,33 +509,70 @@ def iterate(self, *, flat: bool = True) -> Iterator[str | Expr]:
...
+        yield f"!{conversion}"
+    if format_spec:
+        yield ":"
+        for part in format_spec:
+            if isinstance(part, str):
+                # Literal braces must be doubled, as in literal f-string parts.
+                yield part.replace("{", "{{").replace("}", "}}")
+            else:
+                yield from _yield(part, flat=flat, outer_precedence=_OperatorPrecedence.NONE)
+    yield "}"
```
pawamoy, line 537: I gather that we don't need to add a space between the last two `}` in `f"{ {1: 2}}"`, but maybe we should? I suppose people write `f"{ {1: 2} }"` and not `f"{ {1: 2}}"`?

packages/griffelib/src/griffe/_internal/expressions.py
```diff
@@ -686,7 +737,8 @@ class ExprVarKeyword(Expr):
 
     def iterate(self, *, flat: bool = True) -> Iterator[str | Expr]:
         yield "**"
-        yield from _yield(self.value, flat=flat)
+        # Double-starred values are bitwise-or expressions in the grammar, e.g. `f(**a | b)`.
```
pawamoy: I'm not sure to understand these comments. Could you elaborate?

packages/griffelib/src/griffe/_internal/expressions.py
```diff
@@ -1151,8 +1209,8 @@ def iterate(self, *, flat: bool = True) -> Iterator[str | Expr]:
     ExprIfExp: lambda _: _OperatorPrecedence.IF_ELSE,
     ExprNamedExpr: lambda _: _OperatorPrecedence.ASSIGN,
     ExprLambda: lambda _: _OperatorPrecedence.LAMBDA,
-    # NOTE: Ruff categorizes as atomic, but `(a for a in b).c` implies its less than `CALL_ATTRIBUTE`.
-    ExprGeneratorExp: lambda _: _OperatorPrecedence.LAMBDA,
+    # Atomic (like Ruff categorizes it) since it now yields its own parentheses.
```
pawamoy: Do we still need the comment then, now that it matches Ruff's categorization?

packages/griffelib/tests/test_expressions.py
```diff
@@ -88,6 +88,84 @@ def test_expressions(code: str) -> None:
...
+
+
+@pytest.mark.parametrize("context", _expression_contexts)
+def test_expressions_stay_valid_and_equivalent(context: str) -> None:
+    """Stringified expressions must re-parse as valid, semantically-equivalent Python.
+
+    Parameters:
+        context: A code template embedding each expression shape (parametrized).
+    """
+    for shape in _expression_shapes:
```
pawamoy: I think we could use a double-parametrization, like:

```python
@pytest.mark.parametrize("shape", _expression_shapes
@pytest.mark.parametrize("context", _expression_contexts)
def test_expressions_stay_valid_and_equivalent(shape: str, context: str) -> None:
```

It might make things a bit slower, but each the advantage is that each combination is tested in isolation, so all possible failures are displayed after a test run, not just the first ones that made the loop fail.

packages/griffelib/src/griffe/_internal/expressions.py
binggao1230: Kept it as a single element, following your update.

packages/griffelib/src/griffe/_internal/expressions.py
binggao1230: I checked CPython ast.unparse: it normalizes both spellings to the form without a trailing space, so I kept that form here.

packages/griffelib/src/griffe/_internal/expressions.py
binggao1230: Removed both comments; the BIT_OR precedence argument is clearer on its own.

packages/griffelib/src/griffe/_internal/expressions.py
binggao1230: Removed it.

packages/griffelib/tests/test_expressions.py
binggao1230: Applied in f79da51. Each shape/context combination is now a separate parametrized case.

binggao1230:
Updated the PR body to follow the template, including the AI-use checkbox. I also pushed f79da51 for the review suggestions.

pawamoy (approved):
LGTM, thanks a lot!

PR #479, opened 2026-07-16
feat: Support Await expressions

gaoflow:
### For reviewers

- [ ] I did not use AI

### Description of the change

`get_expression()` currently supports every concrete Python expression node except
`ast.Await`, which raises `KeyError`. This adds the missing expression model, builder,
precedence handling, public export, and an AST-equivalence matrix covering Await in its
precedence-sensitive contexts.

### Relevant resources

- https://docs.python.org/3/reference/expressions.html#await-expression
- https://docs.python.org/3/reference/expressions.html#operator-precedence


pawamoy (approved):
Looking good, thanks @gaoflow!

Issue #447, opened 2026-02-24
bug: Stub docstrings are not used for dynamically-analyzed (C extension / SWIG) members


tlambert03:
Hey @pawamoy ... it's me again 

Following up on #116 and #296 — both about [pymmcore](https://github.com/micro-manager/pymmcore), the SWIG-wrapped C extension I keep pestering you about. Those fixes solved the structural problems (wildcard expansion before merge, and wrong alias target paths). The stubs now merge successfully. But there's one remaining gap: `_merge_stubs_docstring` won't overwrite the SWIG-generated docstrings with the proper ones from the .pyi stubs.

The `allow_inspection: false` workaround from #296 was fine when I was rendering pymmcore.CMMCore directly, but I now need `inherited_members: true` on a subclass (`pymmcore_plus.CMMCorePlus`), which requires runtime inspection for MRO resolution — so I can't turn inspection off.

**Problem**

When merging .pyi stubs into a dynamically-inspected C extension module, `_merge_stubs_docstring` never overwrites an existing docstring:

```python
def _merge_stubs_docstring(obj: Object, stubs: Object) -> None:
    if not obj.docstring and stubs.docstring:
        obj.docstring = stubs.docstring
```

This is fine for pure-Python modules where the runtime docstring is the authoritative one. But for SWIG/pybind11/cython extension modules, the runtime "docstrings" are auto-generated C-level signatures like:

```
getCurrentFocusScore(CMMCore self) -> double
```

The real docstrings live in the .pyi stubs.

Since the SWIG methods already have a (not so useful) docstring, the stub docstrings are silently discarded.

There's a second, related issue: For overload-only stub methods (methods where ALL definitions in the .pyi are `@overload` - standard practice for typed C extensions), `_merge_stubs_overloads` correctly attaches the overload annotations to the runtime member, but never applies the overload's docstring to the base function. So even though the overloads carry proper docstrings, the base function retains its SWIG signature string.

**Concrete example**

pymmcore ships an `__init__.pyi` with proper docstrings for all ~290 methods. When griffe loads it with `find_stubs_package=True`, none of the stub docstrings are used — every method renders with its SWIG signature instead.

**Proposed fix**

1. Prefer stub docstrings for dynamically-analyzed objects
    
    Objects loaded via runtime inspection have `analysis = "dynamic"`. When merging stubs, their auto-generated docstrings should yield to the human-authored stub docstrings:
    
    ```diff
[...]

PR #483, opened 2026-08-17
Fix: In packages/griffelib/src/griffe/_internal/merger.py,...

M001N:
## Summary
1) Changed _merge_stubs_docstring to overwrite obj.docstring whenever stubs.docstring exists AND (obj has no docstring OR obj.analysis == 'dynamic'), leaving statically-analyzed genuine docstrings untouched. 2) In _merge_overload_annotations, added a docstring fallback: if the base function has no docstring or is dynamically analyzed, take the docstring from the first overload that has one, before merging annotations.

## Problem
mkdocstrings/griffe issue reference: mkdocstrings/griffe#447

## Root Cause
In packages/griffelib/src/griffe/_internal/merger.py, _merge_stubs_docstring(obj, stubs) used `if not obj.docstring and stubs.docstring` -- it treated 'has any docstring' as sufficient reason to keep the runtime object's docstring, without checking whether that docstring was genuinely authored (obj.analysis == 'static'/None) versus an auto-generated placeholder from runtime inspection (obj.analysis == 'dynamic'), even though griffe's Object model already tracks this via the `analysis: Literal['static','dynamic'] | None` attribute (set by the inspector for dynamically-analyzed members). Separately, _merge_overload_annotations(function, overloads) merged parameter/return type annotations from @overload stub signatures onto the base function but never propagated any overload's docstring onto it.

## Testing
PASS - all 9 tests in test_merger.py pass (3 new + 6 existing); full suite (minus unrelated test_api.py which fails to import due to missing optional 'mkdocstrings' dependency in this environment, and test_git.py which fails due to a missing pytest-git plugin fixture, both pre-existing/unrelated to this change) shows 1742 passed, 22 skipped, 0 failures caused by this change.

## Related Issue
mkdocstrings/griffe#447

### DIFF
--- a/packages/griffelib/src/griffe/_internal/merger.py
+++ b/packages/griffelib/src/griffe/_internal/merger.py
@@ -66,7 +66,7 @@ def _merge_type_alias_stubs(type_alias: TypeAlias, stubs: TypeAlias) -> None:
 
 
 def _merge_stubs_docstring(obj: Object, stubs: Object) -> None:
-    if not obj.docstring and stubs.docstring:
+    if stubs.docstring and (not obj.docstring or getattr(obj, "analysis", None) == "dynamic"):
         obj.docstring = stubs.docstring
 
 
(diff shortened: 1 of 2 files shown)


pawamoy:
Thanks @M001N but we already have a PR for this: https://github.com/mkdocstrings/griffe/pull/481. Let me know if I misunderstood the changes here.

Issue #484, opened 2026-08-20
bug: hit import error while runing test within griffelib


Jazzcort:
### Description of the bug

I encountered these import errors when I tried to package it as RPM for Fedora since within `griffelib` there is no `mkdocstring` and `griffecli` as dependencies. 

I can also bypass these tests by manually skip them while running pytest but I feel like the tests come with the package should be error-free.

### To Reproduce
Download `griffelib` as tarball (tar.gz)
Unzip it and run pytest 

### Full traceback

<details><summary>Full traceback</summary>

```python
====================================================================================== ERRORS ======================================================================================
________________________________________________________________________ ERROR collecting tests/test_api.py ________________________________________________________________________
ImportError while importing test module '/root/rpmbuild/BUILD/python-griffelib-2.2.0-build/griffelib-2.2.0/tests/test_api.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib64/python3.14/importlib/__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/test_api.py:27: in <module>
    from mkdocstrings import Inventory
E   ModuleNotFoundError: No module named 'mkdocstrings'
________________________________________________________________________ ERROR collecting tests/test_git.py ________________________________________________________________________
ImportError while importing test module '/root/rpmbuild/BUILD/python-griffelib-2.2.0-build/griffelib-2.2.0/tests/test_git.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib64/python3.14/importlib/__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/test_git.py:27: in <module>
    from griffe import Module, check, load_git
../BUILDROOT/usr/lib/python3.14/site-packages/griffe/__init__.py:633: in __getattr__
    raise ImportError(f"Please install `griffecli` to use {'griffe.' + attr!r}")
E   ImportError: Please install `griffecli` to use 'griffe.check'
============================================================================= short test summary info ==============================================================================
ERROR tests/test_api.py
ERROR tests/test_git.py
[...]

PR #485, opened 2026-08-20
fix: Skip tests when running without dependencies from the root (griffe)

Jazzcort:
### For reviewers

- [x] I did not use AI

### Description of the change

When running tests within `griffelib` alone as root, importing `griffecli` and `mkdocstrings` at module level causes `ModuleNotFoundError`. Defer these imports and guard them with `pytest.importorskip()` so the tests skip gracefully instead of failing.

Introducing those two dependencies to `griffelib` to just being able to run these tests does not seem like a reasonable choice. Therefore, the best solution I can think of is to gracefully bypass those tests.

I can also bypass those tests by manually skip them while running pytest but I feel like the tests come with the package should be error-free.

### Relevant resources

- #484

### DIFF
--- a/packages/griffelib/pyproject.toml
+++ b/packages/griffelib/pyproject.toml
@@ -41,6 +41,9 @@ classifiers = [
 # to download and inspect packages from PyPI.
 pypi = ["pip>=24.0", "platformdirs>=4.2", "wheel>=0.42"]
 
+[dependency-groups]
+dev = ["pytest>=8.2", "pytest-gitconfig>=0.8.0", "jsonschema>=4.18"]
+
 [tool.uv-dynamic-versioning]
 pattern = '(?P<base>\d+\.\d+\.\d+)'
 
--- a/packages/griffelib/tests/test_api.py
+++ b/packages/griffelib/tests/test_api.py
@@ -24,25 +24,33 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 import pytest
-from mkdocstrings import Inventory
 
 import griffe
-import griffecli
+
+try:
+    import griffecli
+except ModuleNotFoundError:
+    griffecli = None
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator
     from types import ModuleType
 
+    from mkdocstrings import Inventory
 
-TESTED_MODULES = (griffe, griffecli)
+
+TESTED_MODULES = (griffe, griffecli) if griffecli else (griffe,)
 _test_all_modules = pytest.mark.parametrize("tested_module", TESTED_MODULES)
 
 
 @pytest.fixture(name="inventory", scope="module")
 def _fixture_inventory() -> Inventory:
+    pytest.importorskip("mkdocstrings")
     inventory_file = Path(__file__).parent.parent / "site" / "objects.inv"
     if not inventory_file.exists():
         pytest.skip("The objects inventory is not available.")
+    from mkdocstrings import Inventory
+
     with inventory_file.open("rb") as file:
         return Inventory.parse_sphinx(file)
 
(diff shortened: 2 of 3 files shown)


Jazzcort:
more errors showing up. let me look into it.

pawamoy:
Thanks! Looks like the test errors are unrelated to your changes (too old actions?). Let me try to fix them.

pawamoy:
LGTM! Feel free to move out of draft when you're ready, and I'll squash-merge :slightly_smiling_face: Thanks!

Jazzcort:
Hit a few more errors while running test with griffelib as the root. I think `griffe-inherited-docstrings` should not be tested in the `griffelib` test because it actually depends on `griffelib`. I guess this is the artifact that needs to be cleaned up after splitting tests into `griffelib` and `griffecli`. 🤔 

I also added the optional test dependencies for `griffelib` which makes running test within `griffelib` more intuitive. 😁 It's very helpful while running the tests during packaging process.
 
Thank you so much for reviewing and working on this with me! Let me know if you like the changes or not. 😁

packages/griffelib/pyproject.toml
```diff
@@ -40,6 +40,7 @@ classifiers = [
 # The 'pypi' extra provides dependencies needed for the load_pypi functionality
 # to download and inspect packages from PyPI.
 pypi = ["pip>=24.0", "platformdirs>=4.2", "wheel>=0.42"]
+test = ["pytest>=8.2", "pytest-gitconfig>=0.8.0", "jsonschema>=4.18"]
```
pawamoy: Can we set these as dev-dependencies instead? I'm not yet sold on declaring development dependencies as extras.

pawamoy:
Thanks!

>  I guess this is the artifact that needs to be cleaned up after splitting tests into griffelib and griffecli. 🤔

Do you mean there's room for improving the setup here, rather than just excluding the extension?

> Thank you so much for reviewing and working on this with me!

It's only fair! Thanks for packaging the project for Fedora!

packages/griffelib/pyproject.toml
Jazzcort: I'm actually leaning more toward your idea - put the test deps in the dev-dependencies. It's just because I saw so many packages is using `%pyproject_buildrequires -x test`, but the newer "%pyproject_buildrequires -g test" does exist so there shouldn't be any blocker for making them as dev-dependencies. 😁 I'll go ahead and move these dependencies to dev-deps.

Jazzcort:
>Do you mean there's room for improving the setup here, rather than just excluding the extension?

TBH, I think you know more about the module import in python than me. I don't really think I can give you any constructive advices. It's just from my point of view I think `griffe-inherited-docstrings` is extending `griffelib` since it depends on it. `griffelib` shouldn't be testing the functionality of `griffe-inherited-docstrings`. Also, since `griffe-inherited-docstrings` has its own release cycle which may introduce breaking changes. If those breaking changes cause the test to fail in `griffelib`, it's not right in my opinion.


Anyway, it should be good to go. Thanks again! 😁

pawamoy:
We depend on griffe-inherited-docstrings for the documentation of the public API. If we don't load it in our public API tests, I think it doesn't matter because the impacted tests aren't run since I suppose there's no objects.inv in your packaging/testing environment. It's an official extension, we have control over it, so breaking changes are not something to worry about too much :slightly_smiling_face: OK, merging!

pawamoy:
Woops, should have been a `test` type, I'll amend and force-push on main.

Jazzcort:
>We depend on griffe-inherited-docstrings for the documentation of the public API.

I see.

>I think it doesn't matter because the impacted tests aren't run since I suppose there's no objects.inv in your packaging/testing environment.

Yes, you're right. There is no objects.inv in the packaging environment. 😁 

Thanks! It's a pleasure to work with you! 😁

PR #486, opened 2026-08-31
fix: preserve nested f-string format specs

lprnmns:
### For reviewers

- [ ] I did not use AI

### Description of the change

`str(Expr)` added a separating space when a container expression was nested in an f-string format specification. For example, `f'{value:{{1, 2}}}'` was rendered as `f'{value:{ {1, 2}}}'`, which does not preserve the source AST’s format-spec segments. The fix keeps the separator for ordinary replacement fields but suppresses it for nested format fields, and adds a focused regression test.

Validation on Python 3.12:

- `PYTHON_VERSIONS='' ./scripts/make test`
- `PYTHON_VERSIONS='' ./scripts/make check`
- `UV_PROJECT_ENVIRONMENT=.venv uv run --no-sync uv build --all`

### Relevant resources

- Independent reproduction on the current default branch; no exact open issue was found.
- https://github.com/mkdocstrings/griffe/pull/478
- https://docs.python.org/3/library/ast.html#ast.unparse

### DIFF
--- a/packages/griffelib/src/griffe/_internal/expressions.py
+++ b/packages/griffelib/src/griffe/_internal/expressions.py
@@ -531,10 +531,11 @@ def _iterate_format_parts(
     format_spec: Sequence[str | Expr] | None,
     *,
     flat: bool = True,
+    separate_braces: bool = True,
 ) -> Iterator[str | Expr]:
     # Shared by `ExprFormatted` and `ExprInterpolation`.
     yield "{"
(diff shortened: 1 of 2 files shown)


pawamoy (approved):
Thanks, looking good :slightly_smiling_face:

lprnmns:
> Thanks, looking good 🙂

you are welcome sir

PR #487, opened 2026-09-01
fix: Preserve nested tuple parentheses

lprnmns:
### For reviewers

- [ ] I did not use AI

### Description of the change

Keep tuple-context flags local to the immediate AST node. Previously, a subscript or comprehension-target flag could flow through a nested list, call, or comprehension and remove tuple parentheses below it. This changed the rendered AST in cases such as `o[f((x, y))]` and could produce invalid Python for a tuple-valued comprehension nested in a subscript.

Four round-trip regressions cover nested containers, calls, comprehensions, and destructuring targets.

Validation:

- Focused regression: 4 passed
- `PYTHON_VERSIONS='' ./scripts/make format check test` — passed
- Additional deterministic expression matrix — 861 valid cases, 0 failures

### Relevant resources

- Independent reproduction on current `main`; no issue was required to reproduce it.
- Existing PR #480 concerns only top-level named expressions and does not overlap this change.

### DIFF
--- a/packages/griffelib/src/griffe/_internal/expressions.py
+++ b/packages/griffelib/src/griffe/_internal/expressions.py
@@ -1628,6 +1628,11 @@ if sys.version_info >= (3, 14):
 
 
 def _build(node: ast.AST, parent: Module | Class, /, **kwargs: Any) -> Expr:
+    # Tuple context flags apply only to the immediate node. Letting them flow through a nested list, call or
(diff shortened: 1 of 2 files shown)


pawamoy (approved):
Thanks! Found a simplification :slightly_smiling_face: Let me know what you think.

lprnmns:
The simplification looks good to me and keeps the context behavior covered. Thanks for the fixup.

PR #488, opened 2026-09-02
fix: Preserve standalone named expression parentheses

lprnmns:
### For reviewers

- [ ] I did not use AI

### Description of the change

Preserve the required parentheses when a named expression is rendered on its own. Previously, `(value := 1)` became `value := 1`, which cannot be parsed as an expression.

Nested named expressions keep using the existing context-aware renderer, so their output is unchanged.

### Relevant resources

- Reproduced independently on the current `main` branch.
- `PYTHON_VERSIONS='' ./scripts/make check test`
- `.venv/bin/python -m pytest packages/griffelib/tests/test_expressions.py -q` — 918 passed, 16 skipped

### DIFF
--- a/packages/griffelib/src/griffe/_internal/expressions.py
+++ b/packages/griffelib/src/griffe/_internal/expressions.py
@@ -952,6 +952,9 @@ class ExprNamedExpr(Expr):
     value: str | Expr
     """Value."""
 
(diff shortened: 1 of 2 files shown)


pawamoy:
Follow-up of #480.

